PR #10, opened 2018-01-26
Add automatic builds in travis-ci

boxdot:
Also:
* Include Dockerfile of the image used for building and testing.
* Search for gtest package also as module.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,21 @@
+sudo: required
+language: cpp
+
+services:
+  - docker
+
+before_install:
+  - docker pull boxdot/flatdata-dev
+
+env:
+  - C_COMPILER=clang CXX_COMPILER=clang++
+  - C_COMPILER=gcc CXX_COMPILER=g++
+
+script:
+  - |
+    docker run -t \
+      -e C_COMPILER=${C_COMPILER} \
+      -e CXX_COMPILER=${CXX_COMPILER} \
+      -v $(pwd):/src \
+      -w /build \
+      boxdot/flatdata-dev /src/ci/build-and-test.sh
--- a/ci/Dockerfile
+++ b/ci/Dockerfile
@@ -1,0 +1,6 @@
+# CI image for building with gcc and clang
+FROM alpine:latest
+
+RUN apk --no-cache add clang cmake make binutils musl-dev g++ boost-dev gtest gtest-dev gfortran python3-dev && \
+    pip3 install jinja2 pyparsing pandas nose
+
--- a/flatdata-cpp/test/CMakeLists.txt
+++ b/flatdata-cpp/test/CMakeLists.txt
@@ -1,4 +1,4 @@
-find_package(GTest CONFIG QUIET)
+find_package(GTest QUIET)
 find_package(Threads)
 if(NOT GTest_FOUND)
     set(USE_VENDOR_GTEST ON)
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-# flatdata
+# flatdata [![Build Status](https://travis-ci.org/heremaps/flatdata.svg?branch=master)](https://travis-ci.org/heremaps/flatdata)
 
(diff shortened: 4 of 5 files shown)


flatdata-cpp/test/CMakeLists.txt
```diff
@@ -1,4 +1,4 @@
-find_package(GTest CONFIG QUIET)
```
imagovrn: Why this change?

.travis.yml
```diff
@@ -0,0 +1,21 @@
+sudo: required
+language: cpp
+
+services:
+  - docker
+
+before_install:
+  - docker pull boxdot/flatdata-dev
```
imagovrn: Would it be possible to use Dockerfile from source tree? Otherwise, who is able to and how to update the image?

flatdata-cpp/test/CMakeLists.txt
boxdot: To be able to find the package, if it was installed system-wide as a module.

.travis.yml
boxdot: Do you mean to build the image for each test run? This would take too much time since pandas is compiled inside the docker natively. And, most likely, we want to keep the testing environment stable.

If you mean, to build the image on https://hub.docker.com/ from the Dockerfile automatically. Yes, this is possible, however, I don't have permission to add this repository to my dockerhub account.

imagovrn: @boxdot With respect to building image each test run, sounds reasonable. For second question: sorry, I meant, if you suddenly go windsurfing for two weeks, we wouldn't be able to update image at `boxplot/flatdata-dev`, right? Would be cool if there would be `heremaps/flatdata-dev` or otherwise we might want to consider creating `flatdata/dev`. Or am I going overboard?

ci/Dockerfile
```diff
@@ -0,0 +1,6 @@
+# CI image for building with gcc and clang
+FROM alpine:latest
+
+RUN apk --no-cache add clang cmake make binutils musl-dev g++ boost-dev gtest gtest-dev gfortran python3-dev && \
```
imagovrn: Do we really need `musl-dev` and `gfortran`? O_O

boxdot: `musl-dev`: alpine comes with musl libc, we need it since it also contains the runtime object files, needed for creation of elf files.

`gfortran` is needed for compiling pandas (or deps of pandas). Probably, we should consider making pandas optional for building and testing, then we could remove `gfortran`. Also pip3 installations would be obsolete, since the other python packages are part of the alpine distro.

.travis.yml
boxdot: > With respect to building image each test run, sounds reasonable.

Not sure about that. Sounds wasteful, and also dangerous since a failure during the build phase of the image does not necessarily mean that something is wrong with flatdata. Further, an external update of a dependency can break the build. Partially because of that, `Gemfile.lock` (ruby) and `package-lock.json` (npm) were introduced.

> Would be cool if there would be heremaps/flatdata-dev or otherwise we might want to consider creating flatdata/dev. Or am I going overboard?

Not really, I totally agree with your concerns. The best case scenario is when `flatdata-dev` image is built automatically on dockerhub based on the Dockerfile from the source code. A new build should be triggered, when `Dockerfile` is updated in source.

I propose, for now, to go with my intermediate solution. And discuss with `heremaps` a possibility of automatic `heremaps/flatdata-dev` image.

imagovrn: > Not sure about that.
I meant your comment sounds reasonable :)
> I propose, for now, to go with my intermediate solution.
As long as this doesn't become permanent solution :) Could you create an issue for us to remember about it?

ci/Dockerfile
imagovrn: I see. Thanks! +2

boxdot:
As follow-up: #11

PR #12, opened 2018-01-28
Sphinx-powered documentation

imagovrn:
Restructured (and updated) documentation instead of reference.md.

Open questions:
- Generated docs for Python and C++ don't look too good, so might be better to omit them. However long term they could improve :)
- Publishing Sphinx documentation on Github Pages seems quite tricky, so might be worth looking at alternatives.

Feedback welcome.

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,17 +12,17 @@ Flatdata is a library providing data structures for convenient creation, storage
 
 Flatdata helps creating efficient datasets:
 
-* Zero overhead random access.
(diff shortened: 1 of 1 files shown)


imagovrn:
@boxdot if we want this to pass through CI we would need to update docker images. Also, it seems to be quite easy to deploy docs from travis to gh-pages (we can use a separate branch): https://docs.travis-ci.com/user/deployment/pages/

PR #14, opened 2018-01-29
Fixed embarrassing typo in README.md.

imagovrn:


boxdot:
I think you should change a more neutral working for the commit message. Also, it probably makes sense to remove dots from the bullet points.

imagovrn:
@boxdot  done.

PR #15, opened 2018-02-01
Sphinx

imagovrn:
Updated documentation:
- All new docs ported over.
- Optional build.

See https://imagovrn.github.io/flatdata/ as example. If the documentation is in acceptable state we could first push `gh-pages` branch and then submit this PR.


CMakeLists.txt
```diff
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.5)
+cmake_minimum_required(VERSION 3.9)
 project(flatdata)
```
imagovrn: Bumped CMake version for proper Doxygen support.

boxdot:
I built a new docker image with your specs. It is uploaded as `boxdot/flatdata-dev:15` (the tag is this PR number). You could try to use it already in this review, if you modify the ci script. I would propose to hardcode the version of the docker image anyway and update it manually for the time being we don't have automatic builds of images.

VeaaC:
Looks nice, I am in favour of switching

imagovrn:
Hi @boxdot, @VeaaC. I updated the README.md to link correctly to new `.rst` files. And squashed the change to lose the noise in the git log.

boxdot:
Looks good. -- Btw. you can squash when you merge the PR to the master. Adding smaller commits on top makes reviewing much easier since it can be done commit by commit.

PR #20, opened 2018-02-04
Add clang-format specs.

boxdot:


boxdot:
We could. But I think this should never be blocking. It could be a warning with a friendly message saying that it would be nice if the author could format the source code. I don't think we should create hurdles for people to contribute. Some of the developers do not have clang-format installed.

Issue #2, opened 2018-01-25
Add support for enums


boxdot:
It would be nice to have a built-in support for `enum` in flatdata. We could also consider to support a special variant of enums: `bitfield enum`s.

VeaaC:
C++ support was added, but the other languages are lacking it.

PR #21, opened 2018-02-04
Coappearances graph as exhaustive example.

boxdot:
The goal of this example is to show how to convert a nested data format into a flat representation in flatdata (here: json, a sample file is included). Further, it should introduce and use all available data structures in flatdata, hence exhaustive.

When #2 is implemented, this example could be extended to use enums.

I personally think that this example can be also used for an introductory tutorial in flatdata. Especially, the techniques as sentinels, ranges, saving of string in a raw memory blocks, multivectors etc. can be explained easily.


examples/coappearances.cpp
```diff
@@ -13,6 +13,20 @@
 
 namespace co = coappearances;
 
+void
+convert_meta( const picojson::object& obj, co::GraphBuilder builder, std::string& strings )
+{
+    // Since flatdata's mutators are not holding any data, we are creating a vector with a single
+    // element for holding the data.
+    flatdata::Vector< co::Meta > data( 1 );
```
boxdot: Is there a better way to do it right now? -- If not, we should think about a better approach. I find this very contra-intuitive.

One approach could be introducing a method `start_meta` to the builder, which would hide the vector with a single field from the user. `close`  method would then do what `set_meta` is doing now.

examples/coappearances.cpp
```diff
@@ -0,0 +1,327 @@
...
+    auto vertices_data = builder.start_vertices_data( );
+
+    for ( const auto& kv : characters )
+    {
+        const auto& id = kv.first;
+        const auto& data = kv.second.get< picojson::object >( );
+
+        auto character = vertices.grow( );
+        character.name_ref = strings.size( );
+        strings += data.at( "name" ).get< std::string >( ) + '\0';
```
boxdot: This is pretty straight-forward, but still too error-prone. It is very easy to forget to add `+ '\0'` and then the reader would read complete garbage (happened to me several times). What do you think about adding a simple helper class? The interfaces could be similar to `flatdata::Vector`: `grow` and `close`.

examples/coappearances.cpp
```diff
@@ -0,0 +1,327 @@
...
+                auto to_refs = relation.at( "to" ).get< picojson::array >( );
+                assert( to_refs.size( ) == 2 );
+                rel.to_a_ref
+                    = characters_index.at( convert_id( to_refs[ 0 ].get< std::string >( ) ) );
+                rel.to_a_ref
+                    = characters_index.at( convert_id( to_refs[ 1 ].get< std::string >( ) ) );
+            }
+        }
+
+        vertices_data.next_item( );
```
boxdot: This hit me several times already. In `vector` I need first to call `grow` and then start adding data. In the multivector it is the other way around: first, add the data (to the bucket), then call `next_item`. What do you think about unifying the behavior? 

We could even go further and unify the interface by replacing `next_item` by `grow`.

examples/README.md
```diff
@@ -13,3 +13,22 @@ stored.
 
 Two structures are written to the output to show the structure alignment and padding.
 
+Coappearances
+-------------
+
+This examples converts a graph of coappearances from json to flatdata. A graph of coappearances is
```
imagovrn: I think this is a wonderful example, finally ٩(◕‿◕｡)۶

examples/coappearances.cpp
imagovrn: We always wanted to add a `flatdata::Struct< T >` or `flatdata::Object< T >` but there was no time (aka good reason) to. Could do that, if you like (:

examples/coappearances.flatdata
```diff
@@ -0,0 +1,92 @@
...
+    to_a_ref: u32 : 16;
+    to_b_ref: u32 : 16;
+}
+
+struct Chapter {
+    major: u8 : 4;
+    minor: u8 : 7;
+}
+
+// @bound_implicitly( characters: vertices, vertices_data )
```
imagovrn: We never thought of that... "Bound Implicitly"  meant binding two vectors together, as a columnar way to store a single structure.

examples/coappearances.flatdata
```diff
@@ -0,0 +1,92 @@
...
+archive Graph {
+    meta : Meta;
+
+    @explicit_reference( Character.name_ref, strings )
+    vertices : vector< Character >;
```
imagovrn: Why not `characters`? :)

boxdot: I am describing a graph consisting of vertices and edges. This is also a nice way to say vertices are Characters by using types.

examples/coappearances.cpp
```diff
@@ -0,0 +1,327 @@
...
+
+static const char* USAGE = 1 + R"_(
+Usage:
+  coappearances convert <input.json> <output_folder>
+  coappearances read <output_folder>)_";
+
+int
+main( int argc, char const* argv[] )
+{
+    if ( argc < 3 )
```
imagovrn: Might make sense to use `gflags` or similar library to make this leaner. As an example, this can afford extra dependency IMO.

examples/coappearances.cpp
```diff
@@ -0,0 +1,327 @@
...
+        {
+            read( argv[ 2 ] );
+        }
+        else
+        {
+            std::cerr << USAGE << std::endl;
+            return 1;
+        }
+    }
+    catch ( std::runtime_error err )
```
imagovrn: When catching an exception, catch by `const&`. https://stackoverflow.com/questions/2145147/why-catch-an-exception-as-reference-to-const

examples/coappearances.cpp
```diff
@@ -0,0 +1,327 @@
...
+}
+
+uint16_t
+convert_id( const std::string& id )
+{
+    assert( id.size( ) == 2 );
+    return ( static_cast< uint8_t >( id[ 0 ] ) << 8 ) + static_cast< uint8_t >( id[ 1 ] );
+}
+
+using CharactersIndex = std::map< uint16_t /* id */, uint16_t /* ref */ >;
```
imagovrn: This might need a better comment as `id` on one side and `ref` on the other is a little bit confusing.

examples/coappearances.flatdata
```diff
@@ -0,0 +1,92 @@
...
+ *            "a": "character_id",
+ *            "b": "character_id",
+ *            "chapters": [
+ *                "#.##", ...             // # is a digit
+ *            ]
+ *         }
+ *     ]
+ * }
+ */
+namespace coappearances {
```
imagovrn: The schema needs a few comments. Mostly, about the way things refer to each other. Otherwise presence of both "Relations" and "edges" doesn't make it straightforward to understand..

examples/coappearances.cpp
boxdot: Ok, referenced in #23.

examples/coappearances.cpp
boxdot: Yes, makes sense. I also realized that I actually do not need `convert_id` function, since I am not storing ids anyway.

examples/coappearances.cpp
boxdot: Good catch.

examples/coappearances.flatdata
boxdot: Yes, make sense. I also added extended docs to the whole schema.

examples/coappearances/coappearances.cpp
```diff
@@ -0,0 +1,313 @@
...
+Usage:
+  coappearances convert <input.json> <output_folder>
+  coappearances read <output_folder>
+  coappearances -h | --help
+)_";
+
+int
+main( int argc, char const* argv[] )
+{
+    auto args = docopt::docopt( USAGE, {argv + 1, argv + argc} );
```
boxdot: Not sure it makes sense to use something like that for such simple argument parsing.

examples/coappearances.cpp
boxdot: Solved in #25.

examples/coappearances.cpp
boxdot: Used docopt. Seems both to be an overkill. I think I will leave it as is since other examples are also made in the same way. If we want to change it we should make a separate review changing all examples.

examples/coappearances.flatdata
boxdot: Was done in #18.

boxdot:
All comments are fixed. Please review.

examples/coappearances/coappearances.cpp
```diff
@@ -0,0 +1,311 @@
...
+            }
+            else
+            {
+                auto rel = character_data.add< co::BinaryRelation >( );
+                rel.kind_ref = strings.size( );
+                strings += relation.at( "kind" ).get< std::string >( ) + '\0';
+                auto to_refs = relation.at( "to" ).get< picojson::array >( );
+                assert( to_refs.size( ) == 2 );
+                rel.to_a_ref = characters_index.at( to_refs[ 0 ].get< std::string >( ) );
+                rel.to_a_ref = characters_index.at( to_refs[ 1 ].get< std::string >( ) );
```
VeaaC: assigned the same value twice

boxdot: Good catch. Fixed.

PR #25, opened 2018-02-06
Make MultiVector safer/easier to use and more consistent with Vector

VeaaC:
Should now be much safer to use (and in languages like Rust even more so).

Also exposed a bug in our BackwardsCompatibilityTests: They were not filling the last bucket properly (nor testing whether it could actually be read). No harm done, but if we ourselves could not use the old interface properly, it is a clear sign that it needed to be changed.


flatdata-cpp/include/flatdata/MultiVector.h
```diff
@@ -37,14 +42,28 @@ namespace flatdata
...
+    {
+    public:
+        template < typename T >
+        typename T::MutatorType add( );
+
+    private:
+        friend class MultiVector;
+        std::vector< unsigned char >* m_data;
+
+        List( std::vector< unsigned char >* data );
```
imagovrn: `explicit`?

flatdata-cpp/include/flatdata/MultiVector.h
```diff
@@ -37,14 +42,28 @@ namespace flatdata
 template < typename IndexType, typename... Args >
 class MultiVector
 {
+public:
+    /// Class used to fill an individual list
+    class List
```
imagovrn: `ListBuilder` or something similar? Makes `grow`'s signature more clear too.

flatdata-cpp/include/flatdata/MultiVector.h
```diff
@@ -57,54 +76,61 @@ class MultiVector
     ExternalVector< IndexType > m_index;
 
     std::unique_ptr< ResourceHandle > m_handle;
-    size_t m_size = 0;
+    size_t m_size_serialized = 0;
```
imagovrn: `m_bytes_flushed` to be even more precise? Otherwise this managed to confuse me a bit at line 142

flatdata-cpp/include/flatdata/MultiVector.h
```diff
@@ -37,14 +42,28 @@ namespace flatdata
...
+
+        List( std::vector< unsigned char >* data );
+    };
+
 public:
     using View = MultiArrayView< IndexType, Args... >;
 
     MultiVector( ExternalVector< IndexType > index, std::unique_ptr< ResourceHandle > data_handle );
 
-    template < typename T >
```
imagovrn: As this change would break client code, could you please add statement to README.md with something like:

> Flatdata is under active development and our APIs might change. Please expect that when updating from one version to another.

flatdata-cpp/include/flatdata/MultiVector.h
```diff
@@ -113,14 +139,15 @@ template < typename IndexType, typename... Args >
 void
 MultiVector< IndexType, Args... >::add_to_index( )
 {
-    m_index.grow( ).value = m_size;
+    m_index.grow( ).value = m_size_serialized + m_data.size( ) - PADDING_SIZE;
```
imagovrn: In future might it be useful to encapsulate `PADDING_SIZE` somewhere like "padded byte array" as this `-
 PADDING_SIZE` is always there and is quite easy to miss....

flatdata-cpp/test/BackwardCompatibilityTest.cpp
```diff
@@ -137,12 +137,13 @@ std::array< uint8_t, 66 > expected_multivector_data
        "\x00\xff\xff\xff\xff\xef\xbe\xad\xde"          // Payload
        "\x00\x00\x00\x00\x00\x00\x00\x00"};            // Padding
 
-std::array< uint8_t, 37 > expected_multivector_index
-    = {"\x14\x00\x00\x00\x00\x00\x00\x00"    // Index size in bytes
+std::array< uint8_t, 42 > expected_multivector_index
```
imagovrn: As nobody uses us yet (except us), this might be okay, but this comment in README.md should then transform into:

> ....we are changing APIs and binary layout so please expect that....

This will also render all previously built archives inoperable.
The earlier accepted way to do it was to do `mutlivector2` in schema. Lets talk.

flatdata-cpp/test/BackwardCompatibilityTest.cpp
```diff
@@ -266,14 +268,17 @@ TEST( BackwardCompatibilityTest, reading_multivector_resources_layout )
     ASSERT_TRUE( archive.is_open( ) ) << archive.describe( );
 
     auto mv = archive.multivector_resource( );
+    size_t num_data = 0;
```
imagovrn: `expected_items`? `expected_size`?

flatdata-cpp/test/MultiVectorTest.cpp
```diff
@@ -53,6 +53,32 @@ static auto create_view_with_3_items = []( )
     return std::make_pair( std::move( storage ), std::move( view ) );
 };
 
+static const size_t NUM_ITEM_TO_FLUSH = 32 * 1024 * 1024 / 4 + 1024; // enough to flush, and them some
```
imagovrn: `NUM_ITEMS_ENOUGH_TO_FLUSH`?

flatdata-cpp/test/MultiVectorTest.cpp
```diff
@@ -53,6 +53,32 @@ static auto create_view_with_3_items = []( )
...
+
+static auto create_storage_with_enough_items_to_flush = []( )
```
imagovrn: This looks a bit weird considering we always used 'cannonical' function definitions so far. Why lambda?

flatdata-cpp/test/MultiVectorTest.cpp
```diff
@@ -53,6 +53,32 @@ static auto create_view_with_3_items = []( )
...
+{
+    auto storage = MemoryResourceStorage::create( );
+    auto vector = storage->create_multi_vector< TestIndexType48, AStruct, BStructMutator, CStruct >(
```
imagovrn: Why `BStructMutator` and not `BStruct`?

flatdata-cpp/test/MultiVectorTest.cpp
```diff
@@ -251,6 +277,20 @@ TEST( MultiVectorTest, IterateOneTypeElementsRandomlyPlaced )
     ASSERT_TRUE( has_c2 );
 }
 
+TEST( MultiVectorTest, FlushingWhileBuilding )
+{
+    auto view = create_view_with_enough_items_to_flush( );
```
imagovrn: This is not view. I would propose add another variable to avoid `view.second` below

flatdata-cpp/test/MultiVectorTest.cpp
```diff
@@ -251,6 +277,20 @@ TEST( MultiVectorTest, IterateOneTypeElementsRandomlyPlaced )
...
+
+    for ( size_t i = 0; i < NUM_ITEM_TO_FLUSH; i++ )
+    {
+        auto iter = view.second.iterator< CStruct >( i );
+        ASSERT_TRUE( iter.valid( ) ) << "Index " << i;
```
imagovrn: Message: "Expected valid iterator at index: "

flatdata-cpp/test/MultiVectorTest.cpp
```diff
@@ -251,6 +277,20 @@ TEST( MultiVectorTest, IterateOneTypeElementsRandomlyPlaced )
...
+        ASSERT_EQ( i, ( *iter ).value ) << "Index " << i;
```
imagovrn: Message: "Expected: " << i << " at index: " << i << " Actual: " << ...

flatdata-cpp/test/MultiVectorTest.cpp
```diff
@@ -251,6 +277,20 @@ TEST( MultiVectorTest, IterateOneTypeElementsRandomlyPlaced )
...
+        iter++;
+        ASSERT_FALSE( iter.valid( ) ) << "Index " << i;
```
imagovrn: "Expected end of bucket at:" << i ?

flatdata-cpp/test/MultiVectorTest.cpp
imagovrn: Also, this is not "vector" :)

flatdata-cpp/test/BackwardCompatibilityTest.cpp
```diff
@@ -290,21 +295,25 @@ TEST( BackwardCompatibilityTest, reading_multivector_resources_layout )
...
                         [&]( SimpleStruct s )
                         {
                             check_simple_struct( s );
+                            num_data++;
                         },
                         [&]( SignedStruct )
                         {
                             FAIL( );
                         } ) );
+    EXPECT_EQ( 5u, num_data );
```
imagovrn: << "Didn't find enough items......."

flatdata-cpp/include/flatdata/MultiVector.h
VeaaC: Done

flatdata-cpp/include/flatdata/MultiVector.h
VeaaC: Done

flatdata-cpp/include/flatdata/MultiVector.h
VeaaC: Done

flatdata-cpp/include/flatdata/MultiVector.h
VeaaC: Done

flatdata-cpp/test/BackwardCompatibilityTest.cpp
VeaaC: Not necessary, binary layout is the same (test was just creating broken structures without sentinel)

flatdata-cpp/test/BackwardCompatibilityTest.cpp
VeaaC: Done

flatdata-cpp/test/MultiVectorTest.cpp
VeaaC: Done

flatdata-cpp/test/MultiVectorTest.cpp
VeaaC: Done since it's consistent with the rest of the file. Could make sense to change all of them at some point into proper functions

flatdata-cpp/test/MultiVectorTest.cpp
VeaaC: Done. This view does not need to test that Mutator/Accessor are the same for MultiVector

flatdata-cpp/test/MultiVectorTest.cpp
VeaaC: It is consistent with the other tests. Could make sense to change all of them at once at some point

flatdata-cpp/test/MultiVectorTest.cpp
VeaaC: Done automatically by GTest (it does expected + actual + user message). Made the message better, though.

flatdata-cpp/test/MultiVectorTest.cpp
VeaaC: Not require (see above)

flatdata-cpp/test/MultiVectorTest.cpp
VeaaC: See above

flatdata-cpp/test/BackwardCompatibilityTest.cpp
VeaaC: Done

README.md
```diff
@@ -95,6 +95,8 @@ int main() {
 
 For more examples cf. the [examples](examples) directory.
 
+*Note: Until the release APIs are expected to be undergoing (potentially breaking) changes. Binary data layout is stable, though*
```
boxdot: . is missing at the end of the sentence.

flatdata-cpp/include/flatdata/MultiVector.h
```diff
@@ -20,10 +20,15 @@ namespace flatdata
...
+ *        auto list = vector.grow( );
+ *        auto a = list.add< A >( );
+ *        fill_data_a( a );
+ *        auto another_a = list.add< A >( );
+ *        fill_data_a( another_a );
+ *        auto c = list.add< B >( );
+ *        fill_data_b( b );
+ *        auto d = list.add< C >( );
+ *        fill_data_c( c );
  *        vector.next_item( );
```
boxdot: I think this should be `grow`. But as far as I understand, this is not needed at all anymore, correct?

flatdata-cpp/test/MultiVectorTest.cpp
boxdot: Would agree to the change of all of them at once.

README.md
imagovrn: I would move that close to the top of the file to ensure people actually see it. Yeah, and the dot.

flatdata-cpp/test/BackwardCompatibilityTest.cpp
imagovrn: oki

Issue #4, opened 2018-01-25
Add sphinx documentation


imagovrn:
Add autodoc-, doxygen-generated and manually written docs. Publish on GitHub.io

VeaaC:
@imagovrn is this still valid?

imagovrn:
The docs are not yet automatically generated, so somewhat. But I don't think we are going to do it in the nearest future, so we might want to close it for now.

PR #29, opened 2018-02-10
Sphinx documentation generation in travis (#4).

imagovrn:
- Bumped down required CMake version back to 3.5.
- Optional build of docs if CMake 3.9+ is present.
- Building documentation in Travis.
- Updated travis file: variable for docker image,
  mounted docker build, preparation for docs deployment.


.travis.yml
```diff
@@ -4,18 +4,26 @@ language: cpp
...
   - |
     docker run -t \
       -e C_COMPILER=${C_COMPILER} \
       -e CXX_COMPILER=${CXX_COMPILER} \
       -v $(pwd):/src \
+      -v $(pwd)/build:/build \
       -w /build \
-      boxdot/flatdata-dev /src/ci/build-and-test.sh
+      $DOCKER_IMAGE /src/ci/build-and-test.sh
+  - sudo chown -R travis:travis $(pwd)/build
```
boxdot: Are you sure this will work? You are not saying sudo required in travis.

.travis.yml
```diff
@@ -4,18 +4,26 @@ language: cpp
...
+
+before_deploy:
+  - touch $(pwd)/build/docs/html/.nojekyll
```
boxdot: I am confused, why there is `before_deploy` but no `deploy` directive.

.travis.yml
imagovrn: It does. Tried it in my repo + verification passed. Maybe it works by chance? Should I state something extra?

.travis.yml
imagovrn: I kept "before deploy" so the docs preparation is captured. Deploy can't be done just yet. I can also remove it if you like.

imagovrn:
@boxdot @VeaaC should we submit or should I change anything? :)

boxdot:
Any progress on that? Or should we abandon it?

PR #31, opened 2018-02-23
Fix back quote in coappearances example schema for Go implementation

artemnikitin:
I propose this little change because in the current state it will not work with Go implementation.
In case of Go:
```
Raw string literals are character sequences between back quotes, as in `foo`. 
Within the quotes, any character may appear except back quote.
```
https://golang.org/ref/spec#String_literals


artemnikitin:
I created an issue #32

artemnikitin:
@boxdot @imagovrn What are you thinking about it?

PR #34, opened 2018-05-20
Rust generator

boxdot:
Also extend the coappearances examples to show how to use optional subarchives, which do not have to be serialized directly and may be added later.


boxdot:
@imagovrn: I split up the Rust generator templates in files and added tests. I also removed `is_index` method from `Struct` node. Instead, I added a backreference to `multivector` which references the corresponding index type.

Two features are still missing:
* enum types
* optional resources

I would like to add those to the rust generator after I implemented them in https://github.com/boxdot/flatdata-rs.

generator/generators/BaseGenerator.py
```diff
@@ -32,6 +32,8 @@ def render(self, tree):
         env.filters['is_raw_data'] = lambda n: isinstance(n, RawData)
         env.filters['is_archive_resource'] = lambda n: isinstance(n, ArchiveResource)
         env.filters['is_structure'] = lambda n: isinstance(n, Structure)
+        env.filters['is_index'] = lambda n: (isinstance(n, Structure) and
```
VeaaC: is_builtin_structure? We might get more than the simple integer based ones

generator/generators/RustGenerator.py
```diff
@@ -0,0 +1,74 @@
...
+    Archive as ArchiveResource)
+from generator.tree.nodes.trivial import Structure, Constant
+from generator.tree.nodes.archive import Archive
+from .BaseGenerator import BaseGenerator
+
+import re
+
+class RustGenerator(BaseGenerator):
+
+    RESERVED_KEYWORDS = [
```
VeaaC: Nice, we might want to add this for other languages as well :-)

generator/generators/RustGenerator.py
```diff
@@ -0,0 +1,74 @@
...
+        return [Structure, Archive, Constant]
+
+    def _populate_environment(self, env):
+        def _camel_to_snake_case(s):
+            s1 = re.sub('(.)([A-Z][a-z]+)', r'\1_\2', s)
+            return re.sub('([a-z0-9])(A-Z)', r'\1_\2', s1).lower()
+
+        env.filters["camel_to_snake_case"] = _camel_to_snake_case
+
+        def _snake_to_camel_case(s):
```
VeaaC: There are two type of camel case, starting with upper case or not. Maybe be more specific here?

generator/generators/RustGenerator.py
```diff
@@ -0,0 +1,74 @@
...
+        env.filters["rust_doc"] = _rust_doc
+
+        def _escape_rust_keywords(s):
+            if s in self.RESERVED_KEYWORDS:
+                return "{}_".format(s)
+            return s
+
+        env.filters["escape_rust_keywords"] = _escape_rust_keywords
+
+        env.filters['instance_resources'] = lambda ls: [
```
VeaaC: Thought: We could add those to the node then they would available in all generators?

generator/templates/rust/rust.jinja2
```diff
@@ -0,0 +1,29 @@
+{% import "rust/constant.jinja2" as constant %}
+{# {% import "cpp/enumeration.jinja2" as enumeration %} #}
+{% import "rust/structure.jinja2" as structure %}
+{% import "rust/archive.jinja2" as archive %}
+// Do not edit: This code was generated by flatdata's generator.
+
+use flatdata::Struct;
+
+{% macro render_namespaces(root) %}
+{% for module in root.children if module | is_namespace %}
```
VeaaC: What if we do not have any namespace?

generator/generators/BaseGenerator.py
boxdot: In the next line I actually check that this an index.

generator/generators/RustGenerator.py
boxdot: It was quite easy to find a list of keywords for Rust. Also not that hard to find them for C++: http://en.cppreference.com/w/cpp/keyword. I will do same for C++.

generator/generators/RustGenerator.py
boxdot: Renamed to `snake_to_upper_camel_case`.

generator/generators/RustGenerator.py
boxdot: Do you mean "to add them to `BaseGenerator`"?

generator/templates/rust/rust.jinja2
boxdot: Not having a root namespace is an invalid schema. I checked.

generator/generators/BaseGenerator.py
VeaaC: maybe then name it multi_vector_index?

generator/generators/RustGenerator.py
VeaaC: I meant adding them to the syntax tree's node (similar to other node types which can list certain types of children)

boxdot:
All features are implemented except for nested namespaces. I don't see right now an easy way to represent them in Rust. So, I would rather prefer just to flatten them in a single global one.

generator/generators/BaseGenerator.py
boxdot: Fixed

generator/tree/nodes/resources/multivector.py
```diff
@@ -24,13 +25,25 @@ def create(properties, own_schema):
     def _create_references(self):
         return [StructureReference(name=t) for t in self._types]
 
+    @property
```
boxdot: This is a fix. Was missing.

generator/generators/BaseGenerator.py
```diff
@@ -39,6 +39,8 @@ def render(self, tree):
         env.filters['is_bound_resource'] = lambda n: isinstance(n, BoundResource)
         env.filters['is_vector'] = lambda n: isinstance(n, Vector)
         env.filters['is_multivector'] = lambda n: isinstance(n, Multivector)
+        env.filters['is_multivector_index'] = lambda n: (isinstance(n, Structure) and
+            "_builtin.multivector" in SyntaxTree.namespace_path(n))
```
imagovrn: `startswith` instead of `in`?

generator/generators/RustGenerator.py
```diff
@@ -0,0 +1,76 @@
...
+            lines = [
+                re.sub(r'^[ \t]*(/\*\*|/\*|\*/|\*)\s*(.*?)\s*(\*/)?$', r"/// \2", line).strip()
+                for line in s.split('\n')
+            ]
+            start = 0
+            end = len(lines)
+            if lines[0] == "///":
+                start = 1
+            if lines[-1] == "///":
+                end = -1;
```
imagovrn: I am not sure what intended here, but maybe in a same fashion as previous `if` clause it should be `-2`?

generator/generators/RustGenerator.py
```diff
@@ -0,0 +1,76 @@
...
+            return "\n".join(lines[start:end])
```
imagovrn: `_rust_doc` might not work well on empty `lines`

PR #38, opened 2018-06-10
token based schema comparison

fbenkstein:
When opening an archive compare stored schema vs. expected schema with a new
schema_equal comparison function.  This function ignores whitespace and
comments because they do not affect the semantics of the stored binary file.

This allows gradually documenting a schema file (and fixing typos) without
having to recompile all software and binary artifacts.


flatdata-cpp/src/SchemaUtils.re2c.cpp
```diff
@@ -0,0 +1,231 @@
...
+            return t;
+        }
+        <init> punctuation
+        {
+            Token t;
+            t.kind = Token::KIND::PUNCTUATION;
+            t.character = yych;
+            return t;
+        }
+        <init> decimal / ( whitespace | punctuation | nul ) {
```
VeaaC: This should not be required for ignoring whitespaces and comments, should it? And I am not sure that we need to ignore hex vs dec.
Also: negative numbers might be broken

flatdata-cpp/CMakeLists.txt
```diff
@@ -4,12 +4,41 @@ project(flatdata-cpp)
...
+            # readable but slightly slower code.
+            --bit-vectors
+            --start-conditions
+            --input custom
+            -o "${re2c_out}"
+            "${re2c_in}"
+        COMMENT "RE2C ${re2c_in}"
+    )
+
+    list(APPEND FLATDATA_SOURCE "${re2c_out}")
```
VeaaC: How will this avoid a clash with the checked in cpp's symbols?

flatdata-cpp/src/SchemaUtils.re2c.cpp
```diff
@@ -0,0 +1,231 @@
...
+            memset(number_buffer, 0, sizeof(number_buffer));
+            memcpy(number_buffer, start, m_data - start);
+            Token t;
+            t.kind = Token::KIND::NUMBER;
+            t.number = strtoull( number_buffer, nullptr, 10 );
+            return t;
+        }
+        <init> hex / ( whitespace | punctuation | nul ) {
+            memset(number_buffer, 0, sizeof(number_buffer));
+            memcpy(number_buffer, start, m_data - start);
```
VeaaC: What if you exceed the buffer size?

boxdot:
Thank you, @fbenkstein for the contribution. This is a great idea solving an issue which was bugging me for a long time.

Now the comments are removed from the comparison, I wonder, if we could remove them completely from schema files. This would solve the problem for any client, and also make custom comparison implementations not necessary. We would loose some human-addressed information, however I am not sure, this is an issue. What do you think?

VeaaC:
I kind of like the idea of removing comments from the archive itself, since then all languages would get support for ignoring them in a schema comparison automatically.

flatdata-cpp/src/SchemaUtils.re2c.cpp
fbenkstein: This prevents the parser from accepting "123abc" as <Number: 123> <Word: abc>.

fbenkstein:
I think you would also have to normalize whitespace which means basically writing a formatter for .flatdata files.  While possible I don't think it's as easy as it sounds.  Rather the schema could just be transformed to a binary format or even just a hash.

imagovrn:
Nice change! I have one conceptual question though: even though it is nice to have this kind of validation implemented, it can be achieved much easier if comments are not included into the schema, in the generator. Then ignore-whitespace comparison is trivial to implement. Why: if we do it this way, other languages would have to implement similar techniques, in order to be consistent. Is it worth it? Or should we just strip comments in the schema in generator?

Anyway. I think it can only make library better, so if @VeaaC and @boxdot think it's worth it, let's submit it.

VeaaC:
Can you take a look at #44 's flatdata generator, and the schema it generates? Once that is stable enough we could switch the schema stored in the archives.

PR #39, opened 2018-06-11
Add optional archive to coappearances example.

boxdot:
Extend the coappearances example to cover optional subarchives. Ideally, the coappearances example will cover all aspects of flatdata grammar.

The example also shows that optional subarchives can be added later to the main archive.

### DIFF
--- a/examples/coappearances/coappearances.cpp
+++ b/examples/coappearances/coappearances.cpp
@@ -275,16 +275,15 @@ calculate_num_connected_components( co::Graph graph )
 void
 calculate( const char* archive_path )
 {
-    auto graph = co::Graph::open( flatdata::FileResourceStorage::create( archive_path ) );
+    std::shared_ptr< flatdata::FileResourceStorage > storage
+        = flatdata::FileResourceStorage::create( archive_path );
+    auto graph = co::Graph::open( storage );
     if ( !graph )
     {
         std::cerr << graph.describe( ) << std::endl;
         throw std::runtime_error( std::string( "Could not open graph at: " ) + archive_path );
     }
-
-    std::string subarchive_path = std::string( archive_path ) + "/statistics";
-    auto builder = co::StatisticsBuilder::open(
-        flatdata::FileResourceStorage::create( subarchive_path.c_str( ) ) );
+    auto builder = co::GraphBuilder::open( storage ).statistics( );
 
     flatdata::Vector< co::Degree > vertex_degrees( graph.vertices( ).size( ) );
     for ( auto e : graph.edges( ) )
--- a/examples/coappearances/coappearances.cpp
+++ b/examples/coappearances/coappearances.cpp
@@ -212,6 +212,108 @@ convert( const char* json_path, const char* archive_path )
     builder.set_strings( {strings.data( ), strings.size( )} );
 }
 
+// Note: The linear search below is slow for big graphs for obvious reasons, where it is much better
+// to build a lookup table. However, for our small graph this works fine.
+std::vector< uint32_t >
+get_neighbors( co::Graph graph, uint32_t vertex_ref )
+{
+    std::vector< uint32_t > neighbors;
+    for ( auto e : graph.edges( ) )
+    {
+        if ( e.a_ref == vertex_ref )
+        {
+            neighbors.push_back( e.b_ref );
+        }
+        else if ( e.b_ref == vertex_ref )
+        {
+            neighbors.push_back( e.a_ref );
+        }
+    }
+    return neighbors;
+}
+
+size_t
+calculate_num_connected_components( co::Graph graph )
+{
+    size_t num_connected_components = 0;
+
+    std::vector< bool > visited( graph.vertices( ).size( ) );
+    uint32_t num_visited = 0;
+    std::vector< uint32_t > stack;
+
+    while ( num_visited != graph.vertices( ).size( ) )
+    {
+        stack.clear( );
+        // find first non-visited vertex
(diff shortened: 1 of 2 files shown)


examples/coappearances/coappearances.flatdata
```diff
@@ -129,5 +129,26 @@ archive Graph {
 
     // All strings contained in the data separated by '\0'.
     strings: raw_data;
+
+    // Optional archive containing calculated statistics.
+    @optional
+    calculated_data : archive CalculatedData;
```
VeaaC: How about naming it "statistics" with type "Statistics"?

examples/coappearances/coappearances.flatdata
```diff
@@ -129,5 +129,26 @@ archive Graph {
...
+    num_connected_components : u32 : 16;
+}
+
+struct Degree {
+    value : u32 : 16;
+}
+
+archive CalculatedData {
+    invariants : Invariants;
+    vertex_degrees : vector< Degree >;
```
VeaaC: How is this indexed? -> document

examples/coappearances/coappearances.flatdata
```diff
@@ -129,5 +129,26 @@ archive Graph {
 
     // All strings contained in the data separated by '\0'.
     strings: raw_data;
+
+    // Optional archive containing calculated statistics.
+    @optional
+    calculated_data : archive CalculatedData;
+}
+
+struct Invariants {
```
VeaaC: Invariant is kind of a weird name

examples/coappearances/coappearances.cpp
```diff
@@ -212,6 +212,108 @@ convert( const char* json_path, const char* archive_path )
...
+    return num_connected_components;
+}
+
+void
+calculate( const char* archive_path )
+{
+    auto graph = co::Graph::open( flatdata::FileResourceStorage::create( archive_path ) );
+    if ( !graph )
+    {
+        throw std::runtime_error( std::string( "Could not open graph at: " ) + archive_path );
```
VeaaC: also archive.describe?

examples/coappearances/coappearances.cpp
```diff
@@ -212,6 +212,108 @@ convert( const char* json_path, const char* archive_path )
...
+    }
+
+    std::string subarchive_path = std::string( archive_path ) + "/calculated_data";
```
VeaaC: No need to do this, just open a builder for the main archive, and ask for the builder of the subarchive

examples/coappearances/coappearances.cpp
```diff
@@ -212,6 +212,108 @@ convert( const char* json_path, const char* archive_path )
...
+size_t
+calculate_num_connected_components( co::Graph graph )
+{
+    size_t num_connected_components = 0;
+
+    std::vector< bool > visited( graph.vertices( ).size( ) );
+    uint32_t num_visited = 0;
+    std::vector< uint32_t > stack;
+
+    while ( num_visited != graph.vertices( ).size( ) )
```
VeaaC: you can just keep the inner for loop and get rid of the while loop (the one over vertex_ref), as well as num_visited:
Instead of break just do continue

examples/coappearances/coappearances.cpp
```diff
@@ -212,6 +212,108 @@ convert( const char* json_path, const char* archive_path )
...
+                break;
+            }
+        }
+        // depth first search
+        while ( !stack.empty( ) )
+        {
+            uint32_t vertex_ref = stack.back( );
+            stack.pop_back( );
+
+            if ( visited[ vertex_ref ] )
```
VeaaC: Simplification: Mark element when putting on the stack, that way it only ever gets on there once and you do not need to check here

examples/coappearances/coappearances.flatdata
boxdot: Usually, this is a property invariant under a specific transformation; here, graph isomorphism: https://en.wikipedia.org/wiki/Category:Graph_invariants.

examples/coappearances/coappearances.flatdata
boxdot: Indeed, makes sense.

examples/coappearances/coappearances.flatdata
boxdot: Added docs and implicit bound into the parent archive.

examples/coappearances/coappearances.cpp
boxdot: Ah, very good. Added.

examples/coappearances/coappearances.cpp
boxdot: Done.

examples/coappearances/coappearances.cpp
boxdot: Done.

examples/coappearances/coappearances.cpp
```diff
@@ -212,11 +212,110 @@ convert( const char* json_path, const char* archive_path )
...
+{
+    std::shared_ptr< flatdata::FileResourceStorage > storage
+        = flatdata::FileResourceStorage::create( archive_path );
+    auto graph = co::Graph::open( storage );
+    if ( !graph )
+    {
+        std::cerr << graph.describe( ) << std::endl;
+        throw std::runtime_error( std::string( "Could not open graph at: " ) + archive_path );
+    }
+    auto builder = co::GraphBuilder::open( storage ).statistics( );
```
VeaaC: checking for errors here I guess?

PR #41, opened 2018-06-16
Fix non-qualified identifier for explicit references and enums

VeaaC:
Fixes flatadata grammar not allowing fully qualified identifiers for:
- Enum references in struct fields
- explicit_references

### DIFF
--- a/tests/tree/test_syntax_tree_builder.py
+++ b/tests/tree/test_syntax_tree_builder.py
@@ -8,13 +8,14 @@ import sys
 sys.path.insert(0, "..")
 from generator.tree.errors import MissingSymbol
 from generator.tree.builder import SyntaxTreeBuilder
-from generator.tree.nodes.trivial import Namespace, Structure, Field, Constant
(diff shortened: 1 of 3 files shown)


tests/tree/test_syntax_tree_builder.py
```diff
@@ -139,14 +140,26 @@ def test_duplicate_multivector_builtin_types_are_not_produced():
         v1 : multivector< 14, S1 >;
     }
 
+enum Enum1 : u16 {
+ A = 1, B=13, C
+}
+
+struct XXX { e : Enum1; f : .ns.Enum1 : 4; }
```
boxdot: `.` in the beginning is optional, right?

VeaaC: Actually not. The current implementation requires "." to start at the global namespace... similar to C++'s "::". I think we could change this by adding a few more lookup rules (e.g. look into global if lookup fails in your own namespace)

PR #43, opened 2018-06-16
Fix constant handling

VeaaC:
Adds support for:
- signed values
- normalized generator output (decimal always)
- width checking (enough bits for value?)


tests/tree/test_node_tree_builder.py
```diff
@@ -94,13 +94,13 @@ def test_constants_are_placed_on_final_namespace():
 
 def test_constants_are_parsed():
     tree = _build_node_tree("""namespace foo{
-        const i8 foo = 17;
+        const i8 foo = -17;
```
boxdot: Could you please also add a test for negative hex value?

VeaaC: Done

PR #47, opened 2018-06-29
Miscellaneous improvements to flatdata-py.

imagovrn:
+ Adding __doc__ to resources to assist interactive mode.
+ Updated reading routine to be in sync with C++ implementation.
+ Updated vectors' to_data_frame routines to use numpy arrays, reducing memory footprint.
+ Small changes in memory handling in vector - avoiding construction of new mem objects.
+ numpy.dtype support for structures.


imagovrn:
@boxdot I think it's transitive via `pandas`, but explicit is better than implicit. Done.

PR #55, opened 2018-07-11
Fix optional in py

lupax:
Fixed handling of optional in archive.py.
* guard resource open in case of optional decoration in non sub archive schema
* skip resource values in size_in_bytes() in case of not available optional

### DIFF
--- a/flatdata-py/flatdata/archive.py
+++ b/flatdata-py/flatdata/archive.py
@@ -96,16 +96,15 @@ class Archive(object):
 
     def _open_resource(self, name):
         resource_signature = self._schema_validated_resource_signature(name)
-        if not resource_signature:
-            return None
-
-        resource = resource_signature.container.open(storage=self._resource_storage,
-                                                     name=name,
-                                                     initializer=resource_signature.initializer,
-                                                     is_optional=resource_signature.is_optional)
-        if resource:
-            resource.__doc__ = resource_signature.doc
-            return resource
+        if resource_signature:
+            resource = resource_signature.container.open(storage=self._resource_storage,
+                                                         name=name,
+                                                         initializer=resource_signature.initializer,
+                                                         is_optional=resource_signature.is_optional)
+            if resource:
+                resource.__doc__ = resource_signature.doc
+                return resource
+        return None
 
     @staticmethod
     def _is_archive():
--- a/flatdata-py/flatdata/archive.py
+++ b/flatdata-py/flatdata/archive.py
@@ -107,7 +107,6 @@ class Archive(object):
             resource.__doc__ = resource_signature.doc
             return resource
 
-
     @staticmethod
     def _is_archive():
         """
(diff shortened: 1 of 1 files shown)


flatdata-py/flatdata/archive.py
```diff
@@ -79,24 +79,23 @@ def open(cls, storage, name, initializer, is_optional=False):
         return initializer(nested_storage)
 
     def size_in_bytes(self):
-        return sum(
-            self.__getattr__(resource).size_in_bytes()
-            for resource in self._RESOURCES.keys()
-        )
+        return sum( resource_value.size_in_bytes() for resource_value in  ( self.__getattr__(resource)
```
imagovrn: Formatting: please use PEP-8

flatdata-py/flatdata/archive.py
```diff
@@ -79,24 +79,23 @@ def open(cls, storage, name, initializer, is_optional=False):
...
+            for resource in self._RESOURCES.keys() if resource ) if resource_value )
```
imagovrn: `if resource` - is that necessary?

flatdata-py/flatdata/archive.py
```diff
@@ -79,24 +79,23 @@ def open(cls, storage, name, initializer, is_optional=False):
...
         return len(self._RESOURCES)
 
     def _open_resource(self, name):
         resource_signature = self._RESOURCES[name]
-        self._check_non_subarchive_schema(name, resource_signature)
-        resource = resource_signature.container.open(storage=self._resource_storage,
-                                                     name=name,
-                                                     initializer=resource_signature.initializer,
-                                                     is_optional=resource_signature.is_optional)
+        if self._check_non_subarchive_schema(name, resource_signature):
```
imagovrn: This condition will make `resource` on line 95 unassigned.

flatdata-py/flatdata/archive.py
```diff
@@ -109,7 +108,14 @@ def _is_archive():
 
     def _check_non_subarchive_schema(self, name, resource):
         if resource.container._is_archive():
-            return
-        actual_schema = self._resource_storage.get(name + _SCHEMA_EXT).read().decode()
+            return True
+
+        storage = self._resource_storage.get(name + _SCHEMA_EXT, resource.is_optional)
```
imagovrn: I would be happier if this change is done in a separate method. Otherwise it turns method into a `_check_non_subarchive_schema_and_return_true_on_subarchives_or_optional`. A bit too convoluted to my taste.

flatdata-py/flatdata/archive.py
lupax: It's not. I'll fix it.

flatdata-py/flatdata/archive.py
lupax: OK, I'll fix it.

flatdata-py/flatdata/archive.py
lupax: Done. Missed that during merge.

flatdata-py/flatdata/archive.py
lupax: Please see my proposition in the following commit.

flatdata-py/flatdata/archive.py
lupax: Done

PR #57, opened 2018-07-23
Cleanup grammar

VeaaC:
Simplify pyparsing grammar a bit:
- Remove originalText
- Replace setResultsName with call on result (single object), or list (multiple objects)


generator/grammar.py
```diff
@@ -3,9 +3,9 @@
  See the LICENSE file in the root of this project for license details.
 '''
 
-from pyparsing import Word, alphas, alphanums, nums, cppStyleComment, Keyword, Group, Optional, \
-    Or, OneOrMore, originalTextFor, delimitedList, ZeroOrMore, hexnums, Combine, FollowedBy, \
-    ParseException as pyparsingParseException
+from pyparsing import Word, alphas, alphanums, nums, cppStyleComment, \
```
boxdot: Usually, in python you wrap the list after import in ( ). This makes \ obsolete.

VeaaC: Done

PR #59, opened 2018-07-27
Add bitset resources to flatdata-cpp

VeaaC:
* bitset resources are basically a vector<bool>
* supports all vector-like operations: set, stream, view, slice, etc


VeaaC:
Abandon for now.. not sure it's worth the overhead of maintaining this for every language

PR #65, opened 2018-08-21
Return ArrayView when closing ExternalVector.

### LINKED ISSUE #62: When closing ExternalVector return a view to its data.
`ExtenalVector` does not provide indexed access to its data since it may not be yet written to disk. However, after the vector is closed, its data is fully written and constant from that point in time. Often, one wants to access the data after it was closed. So, it would make sense on closing to return an `ArrayView` to its data for convenience.

What do you think, @imagovrn, @VeaaC?
CC: @eikesr.

boxdot:
Impl. detail: ResourceHandle cannot contain a reference to
ResourceStorage, since this would create a circular dependency:
ResourceStorage -> ExternalVector -> ResourceHandle.

Therefore, we provide a black-box read_resource function to
ResourceHandle, which returns a MemoryDescriptor. The latter
is used by ExternalVector to create the corresponding ArrayView.

Resolves #62.

### DIFF
--- a/flatdata-cpp/include/flatdata/ExternalVector.h
+++ b/flatdata-cpp/include/flatdata/ExternalVector.h
@@ -12,6 +12,7 @@
 #include <vector>
 
 #include <boost/noncopyable.hpp>
+#include <boost/optional.hpp>
 
 namespace flatdata
 {
@@ -38,7 +39,7 @@ public:
      */
     ValueType grow( );
 
-    bool close( );
+    boost::optional< ArrayView< T > > close( );
 
 private:
     void flush( );
@@ -65,7 +66,8 @@ ExternalVector< T >::size( ) const
     return m_size;
 }
 
-template< typename T > bool
+template < typename T >
+bool
 ExternalVector< T >::empty( ) const
 {
     return m_size == 0;
@@ -95,11 +97,17 @@ ExternalVector< T >::flush( )
 }
 
 template < typename T >
-bool
+boost::optional< ArrayView< T > >
 ExternalVector< T >::close( )
 {
     flush( );
-    return m_array->close( );
+    boost::optional< MemoryDescriptor > data = m_array->close( );
+    if ( !data )
+    {
+        return boost::none;
+    }
+
+    return ArrayView< T >{data->data( ), data->data( ) + data->size_in_bytes( )};
 }
 
 }  // namespace flatdata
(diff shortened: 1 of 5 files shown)


flatdata-cpp/include/flatdata/internal/ResourceHandle.h
```diff
@@ -59,16 +66,18 @@ ResourceHandle::create( std::shared_ptr< std::ostream > stream )
...
     }
 
+    result->m_resource_reader = std::move( resource_reader );
+
     resource_storage::write_to_stream< resource_storage::size_type >( *result->m_stream, 0 );
     return result;
 }
 
-inline bool
+inline boost::optional< MemoryDescriptor >
```
VeaaC: MemoryDescriptor itself has optional semantics, no need for optional (otherwise user has to double-check: !result || !*result)

flatdata-cpp/include/flatdata/ExternalVector.h
```diff
@@ -95,11 +97,17 @@ ExternalVector< T >::flush( )
 }
 
 template < typename T >
-bool
+boost::optional< ArrayView< T > >
```
VeaaC: ArrayView has optional semantics -> drop optional?

flatdata-cpp/include/flatdata/ExternalVector.h
```diff
@@ -97,17 +103,13 @@ ExternalVector< T >::flush( )
...
     flush( );
-    boost::optional< MemoryDescriptor > data = m_array->close( );
-    if ( !data )
-    {
-        return boost::none;
-    }
-
-    return ArrayView< T >{data->data( ), data->data( ) + data->size_in_bytes( )};
+    MemoryDescriptor data = m_array->close( );
+    return data ? ArrayView< T >{data.data( ), data.data( ) + data.size_in_bytes( )}
```
VeaaC: Actually no need to check, it will just work magically :-)

flatdata-cpp/include/flatdata/internal/ResourceHandle.h
boxdot: Makes sense. Resolved.

flatdata-cpp/include/flatdata/ExternalVector.h
boxdot: Yeah! I rather decided not to look up nullptr arithmetic.

PR #67, opened 2018-08-29
Fix tie-break order of generated entities to match input order...

VeaaC:
... not reverse input order: The topological sort was going through children in reverse order. This caused the normalized schema to change when re-normalized.

### DIFF
--- a/generator/tree/traversal.py
+++ b/generator/tree/traversal.py
@@ -60,7 +60,7 @@ class DfsTraversal(_Traversal):
                 discovered.add(node)
                 stack.append(State(node=node, processed=True))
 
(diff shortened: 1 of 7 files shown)


generator/tree/traversal.py
```diff
@@ -60,7 +60,7 @@ def _iterate(self):
                 discovered.add(node)
                 stack.append(State(node=node, processed=True))
 
-                for child in _Traversal.children(node):
+                for child in reversed(_Traversal.children(node)):
```
boxdot: Could you please add a small comment why this is needed, since it not entirely obvious. Thanks!

PR #70, opened 2018-11-19
Fix example using sentinel's references and crashing

VeaaC:
We should most likely add some tests for it?


examples/coappearances/coappearances.cpp
```diff
@@ -289,7 +289,9 @@ calculate( const char* archive_path )
     auto builder = co::GraphBuilder::open( storage ).statistics( );
 
     flatdata::Vector< co::Degree > vertex_degrees( graph.vertices( ).size( ) );
-    for ( auto e : graph.edges( ) )
+    auto edges = graph.edges( );
+    edges = edges.slice_before( edges.size( ) - 1 );  // skip sentinel
```
boxdot: Thanks for fixing!

PR #72, opened 2018-11-19
Fix ArrayView::slice_after, add tests and mroe convinience functions

VeaaC:
```slice_after( x )``` was actually shortening the length by ``` 2 * x ```

Also added skip (synonym for slice_after), and skip_last (useful for skipping sentinels), as well as front/back.

### DIFF
--- a/flatdata-cpp/test/ArrayViewTest.cpp
+++ b/flatdata-cpp/test/ArrayViewTest.cpp
@@ -11,6 +11,22 @@
 using namespace flatdata;
 using namespace test_structures;
 
+TEST( VectorTest, reading )
+{
+    Vector< AStruct > data( 10 );
+    for ( size_t i = 0; i < 10; i++ )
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: flatdata-cpp/test/ArrayViewTest.cpp [hunk 1]
```diff
@@ -0,0 +1,34 @@
...
+TEST( VectorTest, slicing )
+{
+    Vector< AStruct > data( 10 );
+    for ( size_t i = 0; i < 10; i++ )
+    {
+        data[ i ].value = i;
+    }
+
+    ArrayView< AStruct > view = data;
+    ASSERT_TRUE( view.size( ) == 10 );
```
boxdot: Don't you think we should also check the values?

VeaaC: Done

PR #81, opened 2018-12-02
Add cpp benchmark

VeaaC:
Thoughts?


VeaaC:
@boxdot how about it?

Issue #28, opened 2018-02-09
Make unittesting more stable


boxdot:
Right now, `make test` is not working if `make` was not executed before. This is due to the limitation of downloading the external project and needs to be fixed.

I would like also use the opportunity and a start a discussion about changing the unit testing framework. Our unit tests are fairly simple and do not require mocking, so we don't have use for GMock. On the other hand, GTest needs to be downloaded and built. Which is ok for external dependencies that are really needed for the library, but is cumbersome for unit testing which should just work out of the box. 

I propose to either:

1. Integrate GTest into our source code, or
2. To use something less heavy like [Catch2](https://github.com/catchorg/Catch2).

The latter is a header only single file unit testing library which has proven for me to work already many times.

What do you think, @imagovrn, @VeaaC?

imagovrn:
@boxdot I agree, why not :+1:

PR #83, opened 2018-12-11
Switch testing framework to catch.

boxdot:
For rationale cf. #28.


flatdata-cpp/test/BackwardCompatibilityTest.cpp
```diff
@@ -52,22 +51,20 @@ compare_byte_arrays( const Array& expected,
...
+        INFO( "Difference at position " << i << ". Hexdump: \n" << storage.hexdump( ) );
+        CHECK( actual.data( )[ i ] == static_cast< uint8_t >( expected[ i ] ) );
     }
 }
 
 void
 fill_signed_struct( SignedStructMutator s )
 {
-    EXPECT_EQ( size_t( 10 ), s.size_in_bytes( ) );
+    REQUIRE( s.size_in_bytes( ) == size_t( 10 ) );
```
VeaaC: I guess this will not output both values? Does catch have something that would in case of error? I often find this very convenient.

VeaaC: Actually it seems it does:
```
Example.cpp:9: FAILED:
  REQUIRE( Factorial(0) == 1 )
with expansion:
  0 == 1
```

Nice macro/template magic :+1:

flatdata-cpp/test/BackwardCompatibilityTest.cpp
```diff
@@ -265,62 +262,47 @@ TEST( BackwardCompatibilityTest, reading_multivector_resources_layout )
     storage->assign_value( "multivector_resource_index.schema", multivector_index_schema.c_str( ) );
 
     auto archive = TestMultivector::open( storage );
-    ASSERT_TRUE( archive.is_open( ) ) << archive.describe( );
+    REQUIRE( archive.is_open( ) );
```
VeaaC: INFO with archive.describe()?

flatdata-cpp/test/BackwardCompatibilityTest.cpp
```diff
@@ -340,7 +322,7 @@ TEST( BackwardCompatibilityTest, reading_raw_data_resources_layout )
                            tbi::TestRawData__raw_data_resource__schema__ );
 
     auto archive = TestRawData::open( storage );
-    ASSERT_TRUE( archive.is_open( ) ) << archive.describe( );
```
VeaaC: INFO with archive.describe()?

PR #85, opened 2019-01-12
Separate thousands with an underscore when generating Rust

### LINKED ISSUE #84: Generate long constants with separators
Example:

Instead of 
```
pub const INFINITE_REACH: u32 = 4294967295;
```

generate

```
pub const INFINITE_REACH: u32 = 4_294_967_295;
```

This would remove `clippy` warning in Rust.

gferon:
Resolves #84


generator/generators/RustGenerator.py
```diff
@@ -59,6 +59,14 @@ def _escape_rust_keywords(s):
                 return "{}_".format(s)
             return s
 
+        def _format_numeric_literal(value):
```
boxdot: Since sometimes this function is not doing anything, I would propose to rename it to `try_format_numeric_literal` to avoid confusion, why we are formatting consts of any type as numeric literals.

tests/generators/test_rust_generator.py
```diff
@@ -16,16 +16,23 @@ def test_constants_are_declared_correctly():
...
         const u16 bar = 0x42;
+        // Large constant
+        const u64 foobar = 1000000000;
         }
     """, RustGenerator, """
 /// There is some documentation about foo.
 pub const FOO: i8 = 17;
 """,
-"""
+                           """
```
boxdot: This is intended?

gferon: I'd say it's the same reasoning as `_escape_rust_keywords`, I tried to look in the documentation of `jinja2` but there's no convention. I'd rather keep it as-is, as filters are usually no-op in some case but aren't prefixed with `try_`.

PR #88, opened 2019-02-21
merge boxdot/flatdata-rs

gferon:
TODO:
- package generator in flatdata-rs crate OR have a flatdata-generator crate?
- add example of `build.rs` to have the generator running


boxdot:
What do we do about license and copyright? 

1. flatdata-rs is published with Apache 2.0/MIT dual license. Is this compatible with flatdata's single Apache 2.0 license? Do we have to relicense? Or can we just keep a different license for a subfolder?
2. Copyright is I guess under my username. flatdata's copyright is by HERE. How do we resolve this?

VeaaC:
License: We would have to re-license (all authors need to agree)
Copyright: We need to transfer copyright to HERE (that's what the DCO is about). Actually the copy-right is under the authors, not even you, so it also requires all authors to agree.

Luckily it is only @boxdot @VeaaC and @gferon so if we agree on it it should be easy.

boxdot:
Ok, thanks for clarification. Before agreeing I would need to understand what it actually means to have copyright for something, even the license explicitly states that anybody might copy and use the code even in commercial products.

boxdot:
I checked the legal notion of copyright and license. With this words, I transfer my copyright to HERE Europe B.V. and agree to change the license to a single Apache 2.0 license.

I guess @VeaaC and @gferon, you need to do the same and sign the DCO.

VeaaC:
With this words, I transfer my copyright to HERE Europe B.V. and agree to change the license to a single Apache 2.0 license.

@gferon feel free to mark DCO as "passed", change license and merge

Issue #96, opened 2019-02-23
Create centralized list of flatdata testcases


VeaaC:
Since we now have 4 languages, and each has it's one schemas for tests, I was thinking we should maybe create a central place for testcases:
* for each feature of flatdata a sample schema
* for each corner cases a sample schema
* each with a comment describing what should be tested

That would make it easy to test compliance for new language implementations, and reduce redundancy of the tests.

Opinions @boxdot @imagovrn @gferon ?

boxdot:
This makes totally sense. As you said it makes sense to have a schema for each feature. Then every language can implement an integration test per schema, using generator to generate the corresponding builder/reader. This btw would also allow easily to create a feature matrix.

VeaaC:
Implemented in #99

PR #99, opened 2019-02-25
Add Test case repository and start using it in flatdata/cpp/rust generators

VeaaC:
For #96


VeaaC:
Waiting for #98 to add rust tests

flatdata-py/generator/tests/generators/schemas.py
```diff
@@ -0,0 +1,17 @@
...
+    result = list()
+    basedir = os.path.dirname(__file__)
+    test_dir = os.path.normpath(os.path.join(basedir, '..', '..', '..','..', 'test_cases'))
+    for path, subdirs, files in os.walk(test_dir):
+        for name in files:
+            if os.path.splitext(name)[1] == '.flatdata':
+                relpath = os.path.relpath(path, test_dir)
+                test_relpath = os.path.join(relpath, name)
+                expecation_path = os.path.join(basedir,
+                                               generator + '_expectations',
```
boxdot: @gferon: Could we apply some formatting tool to this file?

flatdata-py/generator/tests/generators/schemas.py
```diff
@@ -0,0 +1,17 @@
...
+                                               os.path.splitext(test_relpath)[0]+'.' + extension)
+                result.append((os.path.join(path, name), expecation_path))
+    assert len(result) == 23, "Did not find expected number of test cases in " + test_dir
```
boxdot: I guess 23 might be a constant at the beginning of the file to make it easily updateble.

flatdata-py/generator/tests/generators/schemas.py
```diff
@@ -0,0 +1,17 @@
+import os
+
+def schemas_and_expectations(generator, extension):
```
boxdot: Could we add a little bit of documentation what are `generator` and `extension` and what is the return type. Unfortunately, it is hard to read python without static types.

flatdata-py/generator/tests/generators/schemas.py
VeaaC: My auto-formatting plugin fave up in this file... I tried to move some expression into variable to improve a bit

flatdata-py/generator/tests/generators/rust_expectations/structs/integers.rs
```diff
@@ -0,0 +1,63 @@
...
+    8,
+    (f, set_f, u64, u64, 0, 64));
+
+define_struct!(
+    I64,
+    RefI64,
+    RefMutI64,
+    schema::structs::I64,
+    8,
+    (f, set_f, i64, i64, 0, 64));
```
boxdot: Just a comment: It looks like we are not generating endline sometimes.

flatdata-py/generator/tests/generators/schemas.py
VeaaC: Done

flatdata-py/generator/tests/generators/rust_expectations/structs/integers.rs
VeaaC: Yeah, our newlines are a bit messed up in the template... sometimes too many, sometimes not enough for the last element of a loop, etc.

I tried to fix for 1h or so, but eventually gave up for now. Will have to do so systematically at some point.

flatdata-py/generator/tests/generators/schemas.py
VeaaC: Done

flatdata-py/generator/tests/generators/rust_expectations/structs/integers.rs
boxdot: No worries, was just a comment.

VeaaC:
Added Rust tests as well and thus this depends on #98 now (too many changes were in that one, so I had to pull it in)

boxdot:
> Added Rust tests as well and thus this depends on #98 now (too many changes were in that one, so I had to pull it in)

Since I was already fine with #98, I think this is fine. This review is in any case a most review, so, from my side we are fine to go. @gferon: Any objections?

PR #105, opened 2019-03-05
Pythonification

gferon:
The goal is to be able to run `pylint` with a good score in CI. Since this is quite a boring PR, I'll try to make the description a bit less boring.

On the top of my head, this makes the following code smells go away:
- :construction: Access to (some) protected members
- :construction_worker_man: Redefining built-in types
- :writing_hand: Non idiomatic namings
- :nut_and_bolt: Classes with only one public method
- :recycle: Cyclic imports

Bonus: Remove C/C++ semicolons :laughing:


flatdata-py/tests/__init__.py
```diff
@@ -0,0 +1,56 @@
+"""Common testing helpers & data"""
```
VeaaC: So much stuff in init files... normally I would never look into those.

gferon: That's part of what's `__init__.py` for, you import `.` and you get the common stuff everywhere if you are part of this module.

PR #110, opened 2019-03-11
Add end-to-end tests for flatdata-rs

VeaaC:
- Including code generation
- Small bugfix for memory storage


flatdata-rs/tests/features/build.rs
```diff
@@ -0,0 +1,34 @@
...
+        if entry.path().extension().map_or(true, |x| x != "flatdata") {
+            continue;
+        }
+
+        let mut result: PathBuf = entry.path().strip_prefix(&test_cases_path).unwrap().into();
+        result.set_extension("rs");
+
+        let filename = out_path.join(&result);
+        std::fs::create_dir_all(filename.parent().unwrap()).expect("Failed to create output dir");
+        let output = std::process::Command::new("python3")
```
gferon: wouldn't it make sense to have all of this in a helper method of the `lib` of `flatdata-rs` so people can build their own `build.rs`?

VeaaC: Yes, completely agree.

flatdata-rs/lib/src/generator.rs
```diff
@@ -0,0 +1,80 @@
...
+            output_dir
+                .join(entry.path().strip_prefix(&schemas_path).unwrap())
+                .with_extension("rs")
+        } else {
+            output_dir
+                .join(entry.path().file_name().unwrap())
+                .with_extension("rs")
+        };
+
+        println!(
```
boxdot: I think you should not print inside build.rs

flatdata-rs/lib/src/generator.rs
```diff
@@ -0,0 +1,80 @@
...
+///
+/// # Examples
+///
+/// `build.rs`
+/// ``` ignore
+/// use std::path::PathBuf;
+///
+/// fn main() {
+///     let out_path = PathBuf::from(std::env::var("OUT_DIR").unwrap());
+///     let test_cases_path = PathBuf::from("assets/schemas");
```
boxdot: We should say that only "*.flatdata" files are consider as schemas.

flatdata-rs/lib/src/generator.rs
```diff
@@ -0,0 +1,80 @@
...
+/// ``` ignore
+/// #![allow(dead_code)]
+///
+/// include!(concat!(env!("OUT_DIR"), "/path/to/my_schema.rs"));
+///
+/// // re-export if desired
+/// pub use my_schema::*;
+/// ```
+pub fn generate(output_dir: &Path, schemas_path: &Path, generator: &Path) {
+    let output_dir = output_dir.canonicalize().unwrap();
```
boxdot: Many unwaps here :/

flatdata-rs/tests/coappearances/Cargo.toml
```diff
@@ -0,0 +1,11 @@
+[package]
+name = "tests_coappearences"
```
boxdot: Should not this package called something with flatdata prefix?

flatdata-rs/lib/src/generator.rs
VeaaC: You can (only special ```cargo:*``` are doing things), but I removed cout printing (cerr will still happen from python generator itself). Input/Output are now part of error and printed in build.rs itself

flatdata-rs/lib/src/generator.rs
VeaaC: Done

flatdata-rs/lib/src/generator.rs
VeaaC: Converted to error

flatdata-rs/tests/coappearances/Cargo.toml
VeaaC: Done

flatdata-rs/tests/coappearances/Cargo.toml
```diff
@@ -0,0 +1,11 @@
+[package]
+name = "flatdata_tests_coappearences"
```
boxdot: Typo: coappearance

flatdata-rs/tests/features/src/archives/multivector.rs
```diff
@@ -0,0 +1,102 @@
...
+                n::RefDataU64Index::S(s) => assert_eq!(s.x(), x as u64),
+                _ => assert!(false, "Found wrong item"),
+            }
+            assert!(iter.next().is_none(), "Too many items");
+        }
+    }
+}
+
+/*
+namespace n{
```
boxdot: This can be removed?

VeaaC: Done

flatdata-rs/tests/coappearances/Cargo.toml
VeaaC: Done

PR #116, opened 2019-03-15
Hide internal symbols from MultiVector/ArrayView API

VeaaC:
Previously:
```MultiVector<osmflat::_builtin::multivector::IndexType40, osmflat::RelationMembers>```
Now:
```MultiVector<osmflat::RelationMembers>```

### DIFF
--- a/flatdata-rs/lib/src/multiarrayview.rs
+++ b/flatdata-rs/lib/src/multiarrayview.rs
@@ -1,6 +1,4 @@
-use crate::archive::{
-    IndexRefFactory, IndexStruct, VariadicRef, VariadicRefFactory, VariadicStruct,
-};
+use crate::archive::{IndexStruct, VariadicRef, VariadicRefFactory, VariadicStruct};
 use crate::arrayview::ArrayView;
 
 use std::fmt;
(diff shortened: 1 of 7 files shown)

### CODE REVIEW

File: flatdata-rs/lib/src/multiarrayview.rs [hunk 1]
```diff
@@ -62,8 +61,9 @@ where
     ///
     /// Panics if index is greater than or equal to `MultiArrayView::len()`.
     pub fn at(&self, index: usize) -> MultiArrayViewItemIter<'a, Ts> {
-        let start = <Idx as IndexStruct>::index(self.index.at(index));
-        let end = <Idx as IndexStruct>::index(self.index.at(index + 1));
+        let start = <<Ts as VariadicStruct<'a>>::Index as IndexStruct>::index(self.index.at(index));
+        let end =
+            <<Ts as VariadicStruct<'a>>::Index as IndexStruct>::index(self.index.at(index + 1));
```
boxdot: Do you think it makes sense to store a part of this statement into a type alias? What clippy says actually?

VeaaC: Doesn't seem possible with the current Rust version (since the alias would depend on some outer paramters)

boxdot: Thanks for trying!

PR #119, opened 2019-03-15
Port remaining generator tests to feature-based tests

### LINKED ISSUE #97: Revamp generator tests
In addition to #96 I was thinking that we might want to simplify the generator tests: Each is testing input schemas vs some generated output, all in code.

Would it make sense to instead use the test schemas from #96 and just store a bunch of expectations as files?

E.g.

test-case ```tests/schemas/signed_enums.flatdata```
expectation ```tests/generator/rust/generated/signed_enums.rs```

Similar to the test it could be that we do not require a 1:1 match, just a "contains" test.

The python test runner would then just enumerate all test schemas, and run against the expectations.

Would that make sense? Is that something you normally do when testing python code, or is the current way preferable @boxdot @imagovrn ?

VeaaC:
Fixes #97


VeaaC:
Also exposes the shortcoming of Go/Python

VeaaC:
Fixes #97

PR #123, opened 2019-03-17
Separate flatdata-py and flatdata-generator at the root of the repository

gferon:
First step to enable publishing `flatdata-generator` on PyPI:
- It is now possible to use `pip3 install flatdata-generator` and get a command-line launcher in `~/.local/bin/flatdata-generator` or in `/usr/local/bin` if installed as root
- Unit tests are totally separate now, which helps for code coverage

Once this is finished, we can have a smart package install in the `build.rs` of `flatdata-rs` (see next PR).

I'm looking for feedback, sorry again for such a huge PR. (First, let's see if this passes CI :crossed_fingers:).


gferon:
Try it out (pip3 install --user flatdata-generator) this is exactly what it already does. Basically you can't install a in-source "launcher", you need to let setuptools generate one since it knows where to look for system or user installed packages.

gferon:
Switching back to `WIP`, I found that both packages are not always playing well with each other in some cases.

VeaaC:
Thanks a lot 👍

Issue #66, opened 2018-08-22
Strategy for closing when destructing ExternalVector (MultiVector).


boxdot:
Behavior right now: When `ExternalVector` is destructed, we have a debug assert, that the user has closed it. Same for `MultiVector`.

In situation, when a user has multiple resources (that need to be closed) encounters an error and wants to make an early return, the resources are destructed, which terminates the program in debug mode. This is obviously not perfect. We should rather try to close in the destructor. The problem is, that closing may fail and we have two possible strategies what to do then.

1. On failure inside close, fail hard. This will terminate the program, since we are failing in a destructor.
2. On failure inside close, do nothing (except for producing a log message).

I tend to use the second approach. Our contract specifies that the user should close resources explicitly. Otherwise, we just try to do it for her, and when we can't do it, we don't want to kill the whole program. This also follows the practice in https://doc.rust-lang.org/std/io/struct.BufReader.html.

@VeaaC, @imagovrn: If you agree, I would implement this.

VeaaC:
C++ and Rust just silently swallow such errors, so we might want to do the same.

PR #127, opened 2019-03-19
Minor flatdata-rs cleanup/formatting

VeaaC:
- Auto-Implement Stream trait
- Remove partially initialized streams by destructing them
- Format includes with merge setting


flatdata-rs/lib/src/filestorage.rs
```diff
@@ -70,7 +70,6 @@ pub struct FileResourceStorage {
 
 impl FileResourceStorage {
     /// Create an empty memory mapped file storage at a given path.
-    #[allow(clippy::new_ret_no_self)]
```
boxdot: this was fixed?

flatdata-rs/lib/src/storage.rs
```diff
@@ -1,24 +1,27 @@
...
+    ops::DerefMut,
+    ptr,
+    rc::Rc,
+    slice, str,
+};
 
 use diff;
 
 pub trait Stream: Write + Seek {}
+impl<T: Write + Seek> Stream for T {}
```
boxdot: Good trick.

flatdata-rs/lib/src/storage.rs
```diff
@@ -293,27 +296,17 @@ impl<'a> ResourceHandle<'a> {
             write_size(0u64, mut_stream.deref_mut())?;
         }
         Ok(Self {
-            stream: Some(stream),
+            stream: stream,
```
boxdot: remove `: stream`.

flatdata-rs/lib/src/storage.rs
```diff
@@ -322,19 +315,13 @@ impl<'a> ResourceHandle<'a> {
 
     /// Close the underlying stream and write the header containing the size in
     /// bytes of written data.
-    pub fn close(&mut self) -> Result<&'a [u8], ResourceStorageError> {
+    pub fn close(self) -> Result<&'a [u8], ResourceStorageError> {
```
boxdot: This is nice, but I don't get to what 'a is bound now.

flatdata-rs/rustfmt.toml
```diff
@@ -1,3 +1,3 @@
-wrap_comments = true
```
boxdot: Are you sure you want to remove it? This is actually very useful.

flatdata-rs/lib/src/filestorage.rs
VeaaC: Yes, two days after this was added :-)

flatdata-rs/lib/src/storage.rs
VeaaC: 'a is a parameter of the class (since the memory is owned by the storage)

flatdata-rs/rustfmt.toml
VeaaC: It broke some long links in the docs... do you know if I can annotate the appropriate places to prevent it from doing that?

flatdata-rs/lib/src/storage.rs
VeaaC: Done

flatdata-rs/rustfmt.toml
VeaaC: #[rustfmt::skip] does not seem to work for the docs at the top of lib.rs

boxdot: Hmm, seems like a bug in rustfmt. Which place did break exactly?

flatdata-rs/lib/src/storage.rs
```diff
@@ -341,12 +341,6 @@ impl<'a> ResourceHandle<'a> {
     }
 }
 
-impl<'a> Drop for ResourceHandle<'a> {
-    fn drop(&mut self) {
-        panic!("Resource not closed");
```
boxdot: Do we really want to remove it? Before we had a check here, if we are not panic'ing already, which seemed to work.

VeaaC: Yeah, this was doing something interesting: When another panic happened, during stack unrolling this was panicking and killing the app.

Also matches what we discussed in #66

boxdot: How about still trying to close the resource in Drop? And if fails, catch the error, log it and not to do anything anymore.

VeaaC: Good idea, will try to do that (also in multivector/externalvector)

VeaaC: Since this seems more complicated (especially in MultiVector it is hard to do without performance impact), I reverted to the previous behaviour for now.

flatdata-rs/rustfmt.toml
VeaaC: Added back and just excluded the problematic file

Issue #73, opened 2018-11-19
Annotations for vectors with sentinels


VeaaC:
Oftentimes flatdata vectors contain a sentinel in the end. It is easy to forget excluding it when iterating.

One idea would be to add an annitation ```@sentinel``` that the generator could use to create two getters for the archive:
* ``` get_xyz() const``` and ``` get_xyz_with_sentinel() const```

What do you think @boxdot @imagovrn ?

boxdot:
I like the idea a lot. I think we should implement it.

VeaaC:
I would not be backwards compatible to add support to language one-by-one, so either we need to:
- Fail schema compilation if not supported
- Implement all language immediately

VeaaC:
I am currently more in favour of implementing another field type (range?), that would automatically mark every vector containing such a struct as having a sentinel, and making it impossible to use such struct in multi_vector.

Would we have use for sentinel annotation beside the range use-case?

boxdot:
Personally, I think that having a sentinel annotation as documentation is already a big plus. Knowing our internal use cases, it is sometimes pretty hard to figure out if a vector is using a sentinel or not. Similarly to implicit bounds and references annotations sentinels annotations would improve readability.

I would still like to hear an opinion from @imagovrn.

VeaaC:
Closing for until. #131 added support for ranges + sentinel.

When we have a use case for sentinels without ranges we can re-open.

PR #131, opened 2019-03-31
range field decoration

VeaaC:
Related-To: #73 

Add range_with_next, supporting the popular "first_xyz_ref" range referencing pattern

Done:
- generator
- flatdata-rs
- flatdata
- tests
- generator checks against mis-usage: no ranges in multivector, single resource, not name clash, etc
- flatdata-cpp

### DIFF
--- a/docs/schema-language.md
+++ b/docs/schema-language.md
@@ -266,12 +266,12 @@ In this case the edges of a node `i` are then retrieved as
 edges.slice(nodes[i].first_edge_ref..nodes[i + 1].first_edge_ref)
 ```
 Additionally the last element of the `nodes` vector is usually a sentinel (only used to retrieve `first_edge_index`).
-To simplify this flatdata offers the `@range_with_next(name_of_range_attribute)` annotation:
+To simplify this flatdata offers the `@range(name_of_range_attribute)` annotation:
 
 ```cpp
 struct Node {
     ...
-    @range_with_next(edges_range)
+    @range(edges_range)
     first_edge_ref : u32;
 }
 ```
--- a/docs/schema-language.md
+++ b/docs/schema-language.md
@@ -237,3 +237,50 @@ namespace N {
 ```
 
 Local paths must be available in the current namespace.
+
+## Index Ranges
+
+When flattening a data model into a flatdata schema one often encounters
+a pattern of storing index ranges as members of consecutive vector items:
+
+```cpp
+struct Node {
+    ...
+    first_edge_ref : u32;
+}
+
+struct Edge {
+    ...
+}
+
+archive Archive {
+    // contains sentinel
+    @explicit_reference( Nodes.first_edge_ref, edges )
+    nodes : vector< Nodes >;
+    edges : vector< Edges >;
+}
+```
+
+In this case the edges of a node `i` are then retrieved as
+```cpp
+edges.slice(nodes[i].first_edge_ref..nodes[i + 1].first_edge_ref)
+```
+Additionally the last element of the `nodes` vector is usually a sentinel (only used to retrieve `first_edge_index`).
+To simplify this flatdata offers the `@range_with_next(name_of_range_attribute)` annotation:
+
+```cpp
+struct Node {
+    ...
+    @range_with_next(edges_range)
+    first_edge_ref : u32;
+}
+```
+
+This will have two effects:
+* Adding `edges_range` attribute exposing range `(nodes[i].first_edge_ref, nodes[i + 1].first_edge_ref)`
+* Hiding the sentinel in views (it still needs to be populated first, though)
+
+Retrieving all edges is now as easy as this:
+```cpp
+edges.slice(nodes[i].edges_range)
+```
--- a/flatdata-generator/flatdata/generator/templates/cpp/structure.jinja2
+++ b/flatdata-generator/flatdata/generator/templates/cpp/structure.jinja2
@@ -1,4 +1,4 @@
-{% set template_header = 'template< template < typename, int, int > class Member >' %}
+{% set template_header = 'template< template < typename, int, int, int > class Member >' %}
 
 {% macro declaration(struct) %}
 {% set template_name = struct.name +'Template' %}
@@ -12,12 +12,16 @@ union {{ template_name }}
 
         {{ field.doc|cpp_doc }}
     {% endif %}
-    using {{ field|typedef_name }} = Member< {{ field.type|cpp_base_type}}, {{ field.offset }}, {{ field.type.width }} >;
+    using {{ field|typedef_name }} = Member< {{ field.type|cpp_base_type}}, {{ field.offset }}, {{ field.type.width }}, {{struct.size_in_bytes}} >;
     {{ field|typedef_name }} {{ field.name }};
+    {% if field.range_with_next %}
+    using {{ field.range_with_next|snake_to_upper_camel_case }}Type = Member< std::pair< {{ field.type|cpp_base_type}}, {{ field.type|cpp_base_type}} >, {{ field.offset }}, {{ field.type.width }}, {{struct.size_in_bytes}} >;
+    {{ field.range_with_next|snake_to_upper_camel_case }}Type {{ field.range_with_next }};
+    {% endif %}
     {% endfor %}
 
     /// Stream type accepted by the class
-    using StreamType = typename Member< uint32_t, 0, 0 >::StreamType;
+    using StreamType = typename Member< uint32_t, 0, 0, 0 >::StreamType;
     /// Mutable structure type
     using MutatorType = {{ template_name }}< flatdata::Writer >;
     /// Immutable structure type
@@ -44,13 +48,15 @@ union {{ template_name }}
     std::string to_string( ) const;
     std::string describe( ) const;
 
+    static constexpr bool IS_OVERLAPPING_WITH_NEXT = {{ struct.has_range_with_next|lower}};
+
     /**
     * Private data member, should not be directly used.
     * Cannot be made private.
     * Please refer to C++ Standard, Chapter 9.2, Paragraph 19.
     * This union has to be kept standard-layout, which different access control prevents.
     */
-    Member< uint32_t, 0, 0 > _data;
+    Member< uint32_t, 0, 0, 0 > _data;
 };
 
 {{ struct.doc|cpp_doc }}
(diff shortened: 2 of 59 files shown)


flatdata-generator/flatdata/generator/templates/rust/structure.jinja2
```diff
@@ -14,5 +14,11 @@ define_struct!(
     {%- if not loop.last %},
     {% endif %}
     {% endfor %}
+    {% for field in struct.fields %}
```
boxdot: Can't this loop be merged with the previous one?

flatdata-generator/flatdata/generator/tree/builder.py
```diff
@@ -152,13 +152,31 @@ def _compute_structure_sizes(root):
...
 
+def _check_ranges(root):
+    # First check that names are unique
+    for field in root.iterate(nodes.Field):
+        name = field.range_with_next
+        if not name:
+            continue
+        for sibling in field.parent.fields:
+            if sibling.name == name:
+                raise InvalidRangeName(name = name)
```
boxdot: I don't think you need `name =` here and below.

flatdata-generator/flatdata/generator/tree/builder.py
```diff
@@ -152,13 +152,31 @@ def _compute_structure_sizes(root):
...
+
+    # Now check that structs with ranges are only used in vectors
+    for reference in root.iterate(StructureReference):
+        if isinstance(reference.node, nodes.Structure) and reference.node.has_range_with_next  \
```
boxdot: wrap the condition with ( ), then you won't need \

flatdata-generator/flatdata/generator/tree/nodes/trivial/structure.py
```diff
@@ -21,6 +21,13 @@ def create(properties, definition):
             result.insert(Field.create(properties=field))
         return result
 
+    @property
+    def has_range_with_next(self):
+        for f in self.fields:
```
boxdot: `any(f for f in self.fields if f.range_with_next)`

flatdata-rs/lib/src/arrayview.rs
```diff
@@ -72,9 +77,10 @@ where
             Bound::Excluded(&idx) => self.data_index(idx + 1),
             Bound::Unbounded => 0,
         };
+        let extra = <T as Struct>::IS_OVERLAPPING_WITH_NEXT as usize;
```
boxdot: 1. I think clippy will complain.
2. Is there actually a better name for `extra`. (If not, is fine :))

flatdata-rs/lib/src/multiarrayview.rs
```diff
@@ -46,12 +46,12 @@ where
     /// the array. An item may be also empty.
     pub fn len(&self) -> usize {
         // last index element is a sentinel
```
boxdot: This comment does not apply anymore, right?

flatdata-rs/lib/src/structs.rs
```diff
@@ -146,12 +151,34 @@ impl<T> VariadicRefFactory for T where T: for<'a> VariadicStruct<'a> {}
...
+
 /// Macro used by generator to define a flatdata struct.
 #[doc(hidden)]
 #[macro_export]
 macro_rules! define_struct {
-    ($factory:ident, $name:ident, $name_mut:ident, $schema:expr, $size_in_bytes:expr
-        $(,($field:ident, $field_setter:ident, $type:path, $primitive_type:tt, $offset:expr, $bit_size:expr))*) =>
+    ($factory:ident, $name:ident, $name_mut:ident, $schema:expr, $size_in_bytes:expr,
+        $(($field:ident, $field_setter:ident, $type:path, $primitive_type:tt, $offset:expr, $bit_size:expr)),*
+        $(,range($range:ident, $range_type:tt, $range_offset:expr, $range_bit_size:expr))*
```
boxdot: Don't you get a double `,` here?

flatdata-rs/tests/coappearances/src/lib.rs
```diff
@@ -70,14 +70,11 @@ fn read_and_validate_coappearances() {
     );
 
     let validate_chapters = |edge_ref, expected| {
-        let e = edges.at(edge_ref);
-        let chapters_start = e.first_chapter_ref() as usize;
-        let chapters_end = edges.at(edge_ref + 1).first_chapter_ref() as usize;
+        let chapters_range = edges.at(edge_ref).chapters_range();
         let e_chapters: Vec<String> = g
             .chapters()
+            .slice(chapters_range.start as usize..chapters_range.end as usize)
```
boxdot: Annoying that there is no map.

flatdata-rs/tests/coappearances/src/lib.rs
```diff
@@ -137,7 +132,7 @@ fn read_and_validate_coappearances() {
     };
 }
 
-fn compare_files(name_a: &path::Path, name_b: &path::Path) -> bool {
+fn compare_files(name_a: &path::Path, name_b: &path::Path) {
```
boxdot: I would rename this function then, since it does not compare anymore.

flatdata-rs/tests/coappearances/src/lib.rs
```diff
@@ -146,15 +141,15 @@ fn compare_files(name_a: &path::Path, name_b: &path::Path) -> bool {
...
+    assert_eq!(buf_a, buf_b);
 }
 
-fn compare_resource(from: &path::PathBuf, to: &path::PathBuf, resource_name: &str) -> bool {
-    compare_files(&from.join(resource_name), &to.join(resource_name))
-        && compare_files(
-            &from.join(format!("{}.schema", resource_name)),
-            &to.join(format!("{}.schema", resource_name)),
-        )
+fn compare_resource(from: &path::PathBuf, to: &path::PathBuf, resource_name: &str) {
```
boxdot: Here the same.

docs/schema-language.md
```diff
@@ -237,3 +237,50 @@ namespace N {
...
+```cpp
+edges.slice(nodes[i].first_edge_ref..nodes[i + 1].first_edge_ref)
+```
+Additionally the last element of the `nodes` vector is usually a sentinel (only used to retrieve `first_edge_index`).
+To simplify this flatdata offers the `@range_with_next(name_of_range_attribute)` annotation:
+
+```cpp
+struct Node {
+    ...
+    @range_with_next(edges_range)
```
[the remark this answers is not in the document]
VeaaC: Agree, name is not very good.
We need a name that reflects that it is a range, but also discerns it from the other type of range:
Type A): data[i].first_ref..data[i+1].first_ref
Type B): data[i].first_ref..data[i].first_ref+data[i].num

flatdata-generator/flatdata/generator/templates/cpp/structure.jinja2
```diff
@@ -44,13 +48,15 @@ union {{ template_name }}
     std::string to_string( ) const;
     std::string describe( ) const;
 
+    static constexpr bool IS_OVERLAPPING_WITH_NEXT = {{ struct.has_range_with_next|lower}};
```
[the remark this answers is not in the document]
VeaaC: It is used in the ArrayView/Vector:
- To hide last element (since it does not have another following one to read data from)

flatdata-generator/flatdata/generator/templates/rust/structure.jinja2
VeaaC: No, the macro definition does not allow mixing different field types (heterogeneous lists), but requires on list type to be followed by the other. Do you know if heterogeneous lists are possible?

flatdata-generator/flatdata/generator/tree/builder.py
VeaaC: If it is not named then it is positional, isn't it?

flatdata-generator/flatdata/generator/tree/builder.py
VeaaC: Wasn't sure what python preferred syntax was, will do

flatdata-rs/lib/src/structs.rs
VeaaC: No: ```,*``` only adds interior commas.

flatdata-rs/tests/coappearances/src/lib.rs
VeaaC: Yes, quite dislike it. How about adding another slice function that does casting for you?
```.slice_as_usize(chapters_range)```

flatdata-generator/flatdata/generator/tree/builder.py
boxdot: Preferred is to use parentheses. I don't remember though where I have this rule from, can't find a reference.

flatdata-generator/flatdata/generator/tree/builder.py
boxdot: Yes, but everything is positional except for `kwargs`.

boxdot:
Thanks @VeaaC! This is a very good extension in my opinion. After we resolved the naming, I think, we can merge.

flatdata-generator/flatdata/generator/templates/rust/structure.jinja2
VeaaC: I would leave this as they are for now, and think about simplifying macros in general in scope of #136

docs/schema-language.md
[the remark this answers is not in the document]
VeaaC: Discussed with @gferon and @imagovrn: They both have no better suggestion and are fine with whatever (requires reading docs anyway)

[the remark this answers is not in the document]
VeaaC: Discussed with @gferon and @imagovrn: They both have no better suggestion and are fine with whatever (requires reading docs anyway)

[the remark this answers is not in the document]
VeaaC: Discussed with @gferon and @imagovrn: They both have no better suggestion and are fine with whatever (requires reading docs anyway)

flatdata-rs/tests/coappearances/src/lib.rs
VeaaC: #135 created to track discussion

Issue #132, opened 2019-04-03
Make extracting strings from raw_data more convinient in Rust


VeaaC:
Typical use cases:
- Retrieve 0-terminated string by index
- Retrieve string by index + size

PR #134, opened 2019-04-09
Add RawData helpers for substrings

VeaaC:
Signed-off-by: Christian Vetter <veaac.fdirct@gmail.com>

This replaces the custom substring methods we have in example / user wrappers

### DIFF
--- a/flatdata-rs/lib/src/rawdata.rs
+++ b/flatdata-rs/lib/src/rawdata.rs
@@ -1,5 +1,5 @@
 /// Exposes blocks of raw data, providing auxiliary functionality like
-/// extracting substrings
+/// extracting substrings.
 #[derive(Debug, Clone, Copy)]
 pub struct RawData<'a> {
     data: &'a [u8],
--- a/flatdata-rs/lib/src/rawdata.rs
+++ b/flatdata-rs/lib/src/rawdata.rs
@@ -33,3 +33,29 @@ impl<'a> RawData<'a> {
         self.data
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn empty() {
+        let data: &[u8] = b"";
+        let raw_data = RawData::new(data);
+        assert_eq!(raw_data.substring(0), Ok(""));
+    }
+
+    #[test]
+    fn last_without_terminator() {
+        let data: &[u8] = b"abc";
+        let raw_data = RawData::new(data);
+        assert_eq!(raw_data.substring(1), Ok("bc"));
+    }
+
+    #[test]
+    fn until_terminator() {
+        let data: &[u8] = b"ab\0c";
+        let raw_data = RawData::new(data);
+        assert_eq!(raw_data.substring(1), Ok("b"));
+    }
+}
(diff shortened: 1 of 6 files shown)


VeaaC:
Fixes #132

flatdata-rs/lib/src/rawdata.rs
```diff
@@ -0,0 +1,35 @@
+/// Exposes blocks of raw data, providing auxiliary functionality like
+/// extracting substrings
```
boxdot: . is missing ;)

flatdata-rs/lib/src/rawdata.rs
```diff
@@ -0,0 +1,35 @@
...
+impl<'a> RawData<'a> {
+    /// Create a new object from raw memory reference
+    pub fn new(data: &'a [u8]) -> Self {
+        Self { data }
+    }
+
+    /// Read a \0 terminated substring starting at specified offset
+    pub fn substring(&self, start: usize) -> Result<&str, std::str::Utf8Error> {
+        let suffix = &self.data[start..];
+        match suffix.iter().position(|&c| c == 0) {
```
boxdot: This is doing 2 scans. First for searching the end and second for checking if it is a well formed utf8 string. Right now (unfortunately) `CStr` is doing the same, however they have a comment that they want to improve this. Probably, we should just use `CStr::from_ptr(..).to_str()` here to benefit from planned changes in the future.

Cf. https://doc.rust-lang.org/src/std/ffi/c_str.rs.html#1159-1165

flatdata-rs/lib/src/rawdata.rs
```diff
@@ -0,0 +1,35 @@
...
+}
+
+impl<'a> RawData<'a> {
+    /// Create a new object from raw memory reference
+    pub fn new(data: &'a [u8]) -> Self {
+        Self { data }
+    }
+
+    /// Read a \0 terminated substring starting at specified offset
+    pub fn substring(&self, start: usize) -> Result<&str, std::str::Utf8Error> {
```
boxdot: Should we also add `substring_unchecked`?

flatdata-rs/lib/src/rawdata.rs
VeaaC: Hmpf, if it were for me they would be removed everywhere :-)

flatdata-rs/lib/src/rawdata.rs
VeaaC: This would fail horribly if the string is not 0-terminated, wouldn't it?

flatdata-rs/lib/src/rawdata.rs
VeaaC: Not for now, we have no unchecked version of any other function (at_unchecked, slice_unchecked, etc). Can however consider adding a bunch of unsafe functions if power-user need those

flatdata-rs/lib/src/rawdata.rs
boxdot: CStr::from_ptr checks for 0-termination at least now.

flatdata-rs/lib/src/rawdata.rs
boxdot: Makes sense.

flatdata-rs/lib/src/rawdata.rs
VeaaC: And our buffer might not contain one. The current version handles that gracefully.

PR #140, opened 2019-06-24
Limit memory consumption of ExternalVector

fermeise:
This prevents the memory consumption for ExternalVector from
being very high for large objects.

### DIFF
--- a/flatdata-cpp/include/flatdata/ExternalVector.h
+++ b/flatdata-cpp/include/flatdata/ExternalVector.h
@@ -83,7 +83,7 @@ template < typename T >
 typename ExternalVector< T >::ValueType
 ExternalVector< T >::grow( )
 {
-    if ( m_data.size( ) > 1024 * 1024 * 32 )
+    if ( m_data.size( ) * sizeof( T ) > 1024 * 1024 * 32 )
     {
         flush( );
     }


flatdata-cpp/include/flatdata/ExternalVector.h
```diff
@@ -83,7 +83,7 @@ template < typename T >
 typename ExternalVector< T >::ValueType
 ExternalVector< T >::grow( )
 {
-    if ( m_data.size( ) > 1024 * 1024 * 32 )
+    if ( m_data.size( ) * sizeof( T ) > 1024 * 1024 * 32 )
```
VeaaC: sizeof( T ) == sizeof( void* ) since all T's are just handles.

What you would need is T::size_in_bytes()... or just m_data.size_in_bytes()

boxdot: I guess you should also fix: https://github.com/heremaps/flatdata/blob/master/flatdata-rs/lib/src/vector.rs#L332

boxdot:
After checking the source more precisely, I see actually that `m_size` is a vector of bytes, so there is no need to multiply with `T::size_in_bytes()`.

PR #124, opened 2019-03-17
Use flatdata-generator when installed in build.rs

gferon:
Since the packages will be soon published to `PyPI`, we can now rely on `flatdata-generator` being installed by `pip`.

This helps addressing issue #106

PR #148, opened 2019-08-15
Bump version to publish on PyPI

gferon:


gferon:
@boxdot and @VeaaC could you create accounts on https://pypi.org so I can add you as maintainer of both `flatdata` and `flatdata-generator` packages?

gferon:
Required before we republish `flatdata-generator` on PyPI in order to verify #124 in CI

PR #149, opened 2019-08-21
Extend raw data's substring

boxdot:


flatdata-rs/lib/Cargo.toml
```diff
@@ -11,6 +11,6 @@ readme = "../README.md"
 edition = "2018"
 
 [dependencies]
-diff = "0.1"
-memmap = "0.6"
-walkdir = "2"
+diff = "0.1.11"
```
VeaaC: Do not pin minor versions? They should be backwards compatible, and that way a user might not get duplicate versions.

boxdot: Actually, it is not pinned: https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#specifying-dependencies-from-cratesio

VeaaC: Ahh good to know

PR #151, opened 2019-08-25
Add force inline to performance critical functions

VeaaC:


boxdot:
I think you removed unintentionally two cpp files.

VeaaC:
Weird, building in-place seems to have deleted those (cmake or make)

PR #156, opened 2019-09-19
Fix C++ warnings

VeaaC:


flatdata-cpp/examples/fibonacci.cpp
```diff
@@ -77,7 +77,7 @@ main( int argc, char const* argv[] )
             return 1;
         }
     }
-    catch ( std::runtime_error err )
+    catch ( std::runtime_error& err )
```
boxdot: This could be even const ref, not?

VeaaC: Done

PR #159, opened 2019-12-09
Make rust generator return error when flatdata-generator fails

stiar:


flatdata-rs/lib/src/generator.rs
```diff
@@ -117,7 +117,7 @@ pub fn generate(
         );
 
         std::fs::create_dir_all(result.parent().unwrap())?;
-        let _ = std::process::Command::new(out_dir.join("bin/flatdata-generator"))
+        let success = std::process::Command::new(out_dir.join("bin/flatdata-generator"))
```
boxdot: Since we imported `std::process` already, you might want to remove this prefix.

flatdata-rs/lib/src/generator.rs
```diff
@@ -127,10 +127,19 @@ pub fn generate(
             .spawn()
             .map_err(|e| GeneratorError::Failure {
                 schema: entry.path().into(),
-                destination: result,
+                destination: result.clone(),
                 error: e,
             })?
-            .wait()?;
+            .wait()?
+            .success();
```
boxdot: You could use `code` here and put it into the error description below. This would provide more information in the error case.

flatdata-rs/lib/src/generator.rs
stiar: Done

flatdata-rs/lib/src/generator.rs
stiar: Done

PR #160, opened 2020-01-18
Move rust macros into python generator

boxdot:
* Python generator now generates directly rust code without using intermediate macros.
* Added a module `test` which contains a flatdata file and a manually generated rust module. The latter is public with hidden docs. The structs and archives from it are used in doc- and unit tests.
* Breaking change in naming: for a flatdata struct `A` we generated now three rust structs `A`, `ARef` and `AMut` instead of `RefA` and `RefMutA`. This is closed to the naming in the Rust's standard lib.

Sorry for a big change. I just don't see how it could have beed split into smaller changes.


flatdata-rs/lib/src/test/test_generated.rs
```diff
@@ -0,0 +1,1310 @@
+// Do not edit: This code was generated by flatdata's generator.
```
VeaaC: Avoid checking this one in? E.g. the feature tests generate code on-demand, which seems to work nicely.

PR #161, opened 2020-01-21
Move out `flatdata::generate` to a separate crate `flatdata-gen`.

boxdot:
This allows to generate Rust code from flatdata schemas in the flatdata library itself. (Needed for #160).

Btw. I just realized that this also allows us to integrate standalone `tests` crates as usual integration tests in to the library crate. However, I won't do it now since this would create conflicts with #160.

### DIFF
--- a/flatdata-rs/Cargo.toml
+++ b/flatdata-rs/Cargo.toml
@@ -1,7 +1,10 @@
 [workspace]
-
 members = [
     "lib",
+    "gen",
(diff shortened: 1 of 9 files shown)


gferon:
LGTM!

boxdot:
I remember we decided not to do this due to some problem, but I don't remember the problem. :| @VeaaC: Do you?

PR #163, opened 2020-02-05
Fix errors in Jinja2

boxdot:
* Make whitespaces in C++ templates more robust.
* Fix trailing comma bug in Jinja.
* The minimal compatible Jinja version is 2.10 (tested all version starting with 2.2).

This change is forward compatible with Jinja2 2.11.1.

### DIFF
--- a/flatdata-generator/tests/generators/cpp_expectations/archives/namespaces.h
+++ b/flatdata-generator/tests/generators/cpp_expectations/archives/namespaces.h
@@ -6,7 +6,7 @@
 #include <iostream>
 #include <iomanip>
 
-namespace n { 
+namespace n {
 
 
 template< template < typename, int, int, int > class Member >
@@ -60,7 +60,7 @@ typedef STemplate< flatdata::Writer > SMutator;
 
 } // namespace n
 
-namespace n { 
+namespace n {
 
 class X : public flatdata::Archive
 {
@@ -125,7 +125,7 @@ private:
 
 } // namespace n
 
-namespace m { 
+namespace m {
 
 
 template< template < typename, int, int, int > class Member >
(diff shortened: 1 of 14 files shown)


flatdata-generator/tests/generators/cpp_expectations/archives/namespaces.h
```diff
@@ -502,7 +503,7 @@ std::string STemplate< Member >::describe( ) const
 }
 } // namespace n
 
-namespace n { 
+namespace n {
```
VeaaC: Pretty mixed style (also previously): Sometimes multiple in one line, sometimes '{' in the same line, sometimes in the next

flatdata-generator/tests/generators/go_expectations/archives/subarchive.go
```diff
@@ -59,8 +59,6 @@ func (v *XArchive) ToString() string {
     buffer := bytes.Buffer{}
     buffer.WriteString(fmt.Sprintf(`{"name": "X", "container_type": "Archive", "size_in_bytes": %d, "resources": [`, v.GetSizeInBytes()))
     buffer.WriteString(v.PayloadRawData.ToString())
-    buffer.WriteString(",")
```
VeaaC: This seems like a behaviour change?

flatdata-generator/tests/generators/cpp_expectations/archives/namespaces.h
boxdot: Should I address it somehow?

flatdata-generator/tests/generators/go_expectations/archives/subarchive.go
boxdot: Yes, this was the trailing comma bug in Go. See go template.

flatdata-generator/tests/generators/cpp_expectations/archives/namespaces.h
VeaaC: No, just an observation.

PR #171, opened 2020-04-18
Generate missing enumeration values to fix enum reading soundness

### LINKED ISSUE #170: Reading enums in Rust can lead to UB in case of incorrect file content
Unlike C/C++ Rust enums are exhaustive: All possible values need to be explicitly mentioned.

That means that casting from the underlying type (e.g. 9 bit integer) to the enumeration can cause UB if the value stored in flatdata does not exist in the enumeration (e.g. corrupted file, etc).

Rust doesn't currently support C-like enums. This means that our options are a bit limited. Some ideas:
* Create enum with an additional (hidden?) field `Unknown`, and check in reader (speed impact?)
* Check enum values every time data is read and panic (speed impact)
* Do not expose as enum, but newtype integer plus constants
* If speed is impacted a `raw` reading function might be needed for fast access

VeaaC:
* Reading enum values that are not listed, but possible due to bit-width can happen, but was not handled.
* We now generate all missing values.
* This requires the user to specify bit-width on the enum, instead of the field (a non-backward compatible change).
* Errors are thrown in case too many missing values would be generated.

Fixes #170


docs/schema-language.md
```diff
@@ -23,15 +23,18 @@ starting with 0), or manually.
...
 
 ```cpp
-enum <Name> : <type> {
+enum <Name> : <type> [ : bits ] {
     <value name> [= value],
     ...
 }
 ```
 
+flatdata will auto-generate names for all missing values as `UNKNOWN_VALUE_X` or `UNKNOWN_VALUE_MINUS_X`.
```
boxdot, line 32: I have to admit, I don't get this. What is `X`? What happens when bits are not not specified?

VeaaC: X is the value, e.g. UNKNOWN_VALUE_1 if 1 hasn't been specified. I will clarify a bit.

If bits are not specified than the full range is assumed, e.g. 8 bits => 256 possible values.
And since all of those could appear in the file (e.g. malicious/corrupted input, etc) they need to be mentioned (otherwise e.g. the transmute done in reading the enum could result in UB)

VeaaC: In the Rust backend this will e.g. generate undocumented values that essentially force the user to add a catch-all to matches (unless the enum is exhausting all the available bits anyway).

I was also considering to an alternative solution: Add checks to the read methods (e.g. replace transmute by a giant match statement), but that seemed to violate the 0-overhead paradigm flatdata has (e.g. the user might not even care for broken enum values for many usecases, e.g just filtering for specific values, etc).

flatdata-generator/flatdata/generator/tree/nodes/trivial/enumeration.py
```diff
@@ -28,9 +31,19 @@ def create(properties, definition):
             result.insert(value_node)
             current_assigned_value += 1
 
-        bits_required = 0
+        # we do not want to genarate too many (exponential) values, so restrict to multiples of input size
+        if len(properties.enum_values) * 2 + 256 < 2 ** result.type.width:
+            # print("Props: " + str(properties.width) + " vs " + str(result.type.width))
```
boxdot, line 36: This print is not needed anymore, rigth?

docs/schema-language.md
```diff
@@ -23,15 +23,18 @@ starting with 0), or manually.
...
 
 ```cpp
-enum <Name> : <type> {
+enum <Name> : <type> [ : bits ] {
     <value name> [= value],
     ...
 }
 ```
 
+Flatdata will auto-generate names for all missing values as `UNKNOWN_VALUE_{X}` or `UNKNOWN_VALUE_MINUS_{X}`, e.g. if no value for `5` is specified, and `5` is possible to represent in the specified number of bits the generator will generate `UNKNOWN_VALUE_5 = 5`. The main reason for this behaviour is, that reading from files is inherently untrustworthy: While the value is not mentioned in the schema, nothing prevents a malicious entity from writing it.
```
boxdot, line 32: Thank you for extending the explantion!

flatdata-generator/tests/generators/cpp_expectations/enums/implicit_values.h
```diff
@@ -9,5 +9,8 @@ enum class Enum1 : uint16_t
     // = 1
     VALUE_4 = 1,
     // = 2
-    VALUE_5 = 2
-};
\ No newline at end of file
+    VALUE_5 = 2,
+    UNKNOWN_VALUE_5 = 5,
```
boxdot, line 13: One think which I still don't quite get. Wouldn't it be better just to generate a single UNKNOWN variant and map all unknown value to it?

flatdata-generator/tests/generators/rust_expectations/enums/values.rs
```diff
@@ -7,6 +7,506 @@ pub enum EnumI8 {
     FooI8NegHex = -127,
     FooI8PosHex = 126,
     FooI8OneHex = 1,
+    #[doc(hidden)]
+    UnknownValueMinus126 = -126,
```
boxdot, line 11: I am not sure what is the benefit of generating all these value. In the end the user can't match them all, and they even do not know what to match against since the names are hidden in the docs.

flatdata-generator/tests/generators/cpp_expectations/enums/implicit_values.h
VeaaC: This is done for performance reasons: This way reading from the field is for-free: Just a transmute without any branches

flatdata-generator/flatdata/generator/tree/nodes/trivial/enumeration.py
VeaaC: Removed

flatdata-generator/tests/generators/rust_expectations/enums/values.rs
VeaaC: Without these some compilers will have undefined behaviour when transmuting these values unchecked.

This is essentially similar to the current practise of non-exhaustive Rust enums: https://github.com/rust-lang/rfcs/blob/master/text/2008-non-exhaustive.md

flatdata-generator/tests/generators/cpp_expectations/enums/implicit_values.h
boxdot: Thanks for explaining. Now, I see it better. Following the 0-overhead rule we have to do this.

PR #172, opened 2020-04-19
Add @const( <name> ) annotation for fields to indicate which constant is used where

VeaaC:
Any constant references this way will only be included in the schema of resources referencing it

### DIFF
--- a/docs/schema-language.md
+++ b/docs/schema-language.md
@@ -14,6 +14,14 @@ Flatdata supports the following primitive types:
 -  ``i64`` - signed 64-bit wide type
 -  ``u64`` - unsigned 64-bit wide type
 
+## Constants
+
+Flatdata supports defining constants of basic types:
+
+```cpp
+const <type> <name> = <value>;
+```
+
 ## Enumerations
 
 Flatdata supports adding enumeration over basic types. Each enumeration
@@ -23,7 +31,7 @@ starting with 0), or manually.
 Each enumeration is defined as follows:
 
 ```cpp
-enum <Name> : <type> [ : bits ] {
+enum <name> : <type> [ : bits ] {
     <value name> [= value],
     ...
 }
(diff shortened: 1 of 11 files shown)


docs/schema-language.md
```diff
@@ -152,6 +160,23 @@ and create reference edges, while other generators mostly support only
...
+
+```cpp
+const u32 MY_CONST = 10;
+struct MyStruct {
+    @const( MY_CONST )
+    my_value : u32 : 16;
+}
+```
+
+Note: If a constant is not referenced anywhere, flatdat will assume that
```
boxdot, line 176: typo

docs/schema-language.md
```diff
@@ -152,6 +160,23 @@ and create reference edges, while other generators mostly support only
...
 
+## Constant Referencing
+
+``@const( <name> )`` can be added to fields of a structure to indicate
+in which locations a constant can appear, e.g.:
+
+```cpp
+const u32 MY_CONST = 10;
+struct MyStruct {
+    @const( MY_CONST )
```
boxdot, line 171: This is only for documentation purposes (and dot file) atm, right?

flatdata-generator/flatdata/generator/tree/builder.py
```diff
@@ -8,7 +8,7 @@
 import flatdata.generator.tree.nodes.trivial as nodes
 from flatdata.generator.grammar import flatdata_grammar
 from flatdata.generator.tree.errors import (
-    InvalidEnumWidthError, InvalidRangeName, InvalidRangeReference)
+    InvalidEnumWidthError, InvalidRangeName, InvalidRangeReference, InvalidConstReference, InvalidConstValueReference)
```
boxdot, line 11: The line might be a little bit too long

docs/schema-language.md
VeaaC: It also affects some schemas:
The constants are only serialized into the schema of resources using it (instead of the full archive)

flatdata-generator/flatdata/generator/tree/builder.py
VeaaC: Fixed

boxdot:
There are still some conflict with master.

PR #176, opened 2020-05-02
Move schemas to the end of generated rust code (fixes #168)

### LINKED ISSUE #168: Move generated schemas down in generated Rust code.
Schema definitions are not interestring to read and are hidden in the docs anyway. They should down in the file.

VeaaC:
Also remove unused struct schemas

Fixes #168


VeaaC:
Depends on #175

Issue #173, opened 2020-04-19
Add @optional( <name> ) annotation for optional fields


VeaaC:
a common pattern used in flatdata is too use a fixed value to indicate a missing field, e.g.

```cpp
const u32 INVALID_INDEX = 0xFFFFFFFF;
struct MyStruct {
    @const( INVALID_INDEX )
     index : u32;
}
```

I propose making this a use case that we should support properly:
* add `@optional( <const_name> )` annotation
* generate `Option<T>/boost::optional/whatever` getters/setters
* Allow fast access to raw data via special getter (e.g. `field_name_raw()`?)

VeaaC:
Duplicates #80

PR #178, opened 2020-05-04
Add @optional annotation

### LINKED ISSUE #177: C++: Struct's to_string method does not handle signed types
Every struct member is converted to uint64_t

VeaaC:
* `@optional( <name> )` field annotation to mark null values of fields
* C++ and Rust support

Fixes #177 and #173


VeaaC:
Depends on #176 and #175

PR #184, opened 2020-05-15
Bump versions in preparation of release

VeaaC:


flatdata-rs/lib/src/generator.rs
```diff
@@ -85,7 +85,7 @@ pub fn generate(
         path
     } else {
         eprintln!("installing flatdata-generator from PyPI");
-        PathBuf::from("flatdata-generator")
+        PathBuf::from("flatdata-generator=0.4")
```
gferon, line 88: you want `==0.4.*` to be super explicit

VeaaC: Done

Issue #188, opened 2020-05-25
Archive::describe does not describe sub-archives


VeaaC:
Mostly a C++ issue (Rust is actually not able to describe failure to open archives at all)

PR #189, opened 2020-05-29
Fix Archive::describe to include sub-archives (#188)

InfinityByTen:
Added to template adding describe to optional resources.
Prepended a newline for readability.

### DIFF
--- a/flatdata-cpp/include/flatdata/Archive.h
+++ b/flatdata-cpp/include/flatdata/Archive.h
@@ -151,7 +151,8 @@ Archive::describe_resource( std::ostream& stream,
 {
     auto initialized = static_cast< bool >( resource );
     describe_impl( stream, name, true, initialized ? static_cast< bool >( *resource ) : false,
-                   initialized ? resource->describe( ).c_str( ) : "N/A", too_large );
+                   initialized ? ( std::string( "\n" ) + resource->describe( ) ).c_str( ) : "N/A",
+                   too_large );
 }
 
 template < typename ResourceType >
--- a/flatdata-cpp/include/flatdata/Archive.h
+++ b/flatdata-cpp/include/flatdata/Archive.h
@@ -150,9 +150,19 @@ Archive::describe_resource( std::ostream& stream,
                             bool too_large )
 {
     auto initialized = static_cast< bool >( resource );
+    std::string description;
+    if ( initialized )
+    {
+        description = std::is_base_of< Archive, ResourceType >::value
+                          ? ( std::string( "\n" ) + resource->describe( ) )
+                          : resource->describe( );
+    }
+    else
+    {
+        description = "N/A";
+    }
     describe_impl( stream, name, true, initialized ? static_cast< bool >( *resource ) : false,
-                   initialized ? ( std::string( "\n" ) + resource->describe( ) ).c_str( ) : "N/A",
-                   too_large );
+                   description.c_str( ), too_large );
 }
 
 template < typename ResourceType >
--- a/flatdata-cpp/include/flatdata/Archive.h
+++ b/flatdata-cpp/include/flatdata/Archive.h
@@ -136,26 +136,14 @@ private:
 
 template < typename ResourceType >
 std::string
-get_description( const ResourceType& resource,
-                 bool is_archive,
-                 bool initialized,
-                 size_t nest_level )
+get_description( const ResourceType& resource, bool is_archive, size_t nest_level )
 {
     std::string description;
     if ( is_archive )
     {
         ++nest_level;
     }
-
-    if ( initialized )
-    {
-        description = resource.describe( nest_level );
-    }
-    else
-    {
-        auto dummy = new ResourceType( );
-        description = dummy->describe( nest_level );
-    }
+    description = resource.describe( nest_level );
     return description;
 }
 
@@ -171,8 +159,8 @@ Archive::describe_resource( size_t nest_level,
     const bool is_archive = std::is_base_of< Archive, ResourceType >::value;
 
     describe_impl( stream, name, false, initialized,
-                   get_description( resource, is_archive, initialized, nest_level ).c_str( ),
-                   is_archive, too_large, nest_level );
+                   get_description( resource, is_archive, nest_level ).c_str( ), is_archive,
+                   too_large, nest_level );
 }
 
 template < typename ResourceType >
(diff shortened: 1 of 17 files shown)


VeaaC:
Can you add some unit tests as well (GeneratedArchiveTests)?:
* test that a failed sub-archive can be described
* test that a failed optional sub-archive can be described

flatdata-cpp/test/GeneratedArchiveTest.cpp
```diff
@@ -330,6 +330,72 @@ TEMPLATE_TEST_CASE_METHOD( Fixture,
     REQUIRE_FALSE( outer.inner( ).is_open( ) );
 }
 
+TEMPLATE_TEST_CASE_METHOD( Fixture,
+                           "Uninitialized sub-archive is descried",
```
VeaaC, line 334: desried -> described

flatdata-cpp/include/flatdata/Archive.h
```diff
@@ -150,8 +150,19 @@ Archive::describe_resource( std::ostream& stream,
                             bool too_large )
 {
     auto initialized = static_cast< bool >( resource );
+    std::string description;
+    if ( initialized )
+    {
+        description = std::is_base_of< Archive, ResourceType >::value
```
VeaaC, line 156: What about:
* describing resourcing is sub-archives that failed to initialize? How does that look like now?
* describing resources in sub-archives that successfully loaded, how does that look like now?

I guess it might need to look like a tree-structure, e.g.:
```
resource1 : ...
archive_resource1: ...
    sub_resource1: ...
    sub_resource2: ...
```

InfinityByTen: * Uninitialized is reported as an error only in mandatory case. But then it doesn't stop us from describing the `Archive` child resources instead of keeping it "N/A". Will adapt it.

* If everything is initialized, then then describe works but there is no indentation and everything is flattened out (it's flatdata after all). But I'll try to see if there's a nice way to get indentation.

flatdata-cpp/test/GeneratedArchiveTest.cpp
InfinityByTen: Ah sure, will fix.

flatdata-cpp/include/flatdata/Archive.h
InfinityByTen: Would you suggest a tree structure with complete descriptions like so? :
flatdata_tree_describe

VeaaC: Looks a bit hacky... how about having a single combined list, but full resource names:

```
outermost
outer
outer.outer1
outer.outer2
outer.inner
outer.inner.inner
```

That could require passing down the path-prefix, and being able to disable headers/footers during recursive calls

InfinityByTen: Well, first versions are mostly hacky. Anyhow, I've reworked it a bit based on your inputs. I would like to retain the tree structure and somehow include which ones are archives and which are plain structs. Here's how the latest version looks:
describe_reworked

Let me know what you think.

VeaaC: Looks good!

InfinityByTen:
I seem to often run into this issue of tests failing on CI. This time it happened because I just had forgotten where to run the tests from. Should we have a separate task to put into a README how to run tests locally?

VeaaC:
> I seem to often run into this issue of tests failing on CI. This time it happened because I just had forgotten where to run the tests from. Should we have a separate task to put into a README how to run tests locally?

That makes sense: Some `How to contribute` section with advice how to test the individual components:
* `python -m nose` for generator tests
* how to manually run generator for a test to get a new reference, e.g. `./generator.py -s ../test_cases/archives/multivector.flatdata -g rust -O tests/generators/rust_expectations/archives/multivector.rs.1`
* how to update the checked in generated rust test code, e.g.: `../flatdata-generator/generator.py -s lib/src/test/test.flatdata -g rust -O lib/src/test/test_generated.rs && sed -i 's/flatdata::/crate::/g' ../flatdata-rs/lib/src/test/test_generated.rs`
* etc

Do you want to create a new issue for that? And would you like to fill that yourself, or have it filled by regular contributors?

InfinityByTen:
> Do you want to create a new issue for that? And would you like to fill that yourself, or have it filled by regular contributors?

I can create an issue as a starting point, definitely. But I guess I don't have enough insight as to what all it could include. Like I literally had no idea about half the things you mentioned.

flatdata-cpp/include/flatdata/Archive.h
```diff
@@ -130,28 +134,64 @@ class Archive
...
+        ++nest_level;
+    }
+
+    if ( initialized )
+    {
+        description = resource.describe( nest_level );
+    }
+    else
+    {
+        auto dummy = new ResourceType( );
```
VeaaC, line 156: You are leaking memory here, just create the object on the stack?

flatdata-cpp/include/flatdata/Archive.h
```diff
@@ -130,28 +134,64 @@ class Archive
...
                             const char* name,
                             const boost::optional< ResourceType >& resource,
                             bool too_large )
 {
-    auto initialized = static_cast< bool >( resource );
+    const auto initialized = static_cast< bool >( resource );
+    const bool is_archive = std::is_base_of< Archive, ResourceType >::value;
+
+    const ResourceType ref = initialized ? *resource                  // valid ref
+                                         : *( new ResourceType( ) );  // ref to dummy, not used
```
VeaaC, line 190: You are leaking memory here, just create the object on the stack?

flatdata-cpp/include/flatdata/ArrayView.h
```diff
@@ -44,7 +44,7 @@ class ArrayView
 
     size_t size_in_bytes( ) const;
     size_t size( ) const;
-    std::string describe( ) const;
+    std::string describe( size_t unused = 0u ) const;
```
VeaaC, line 47: give unused arguments their normal name, so users know what they are

flatdata-cpp/include/flatdata/MemoryDescriptor.h
```diff
@@ -31,10 +31,17 @@ struct MemoryDescriptor
     }
 
     std::string
-    describe( ) const
+    describe( size_t unused = 0 ) const
```
VeaaC, line 34: size_t /*name_of_variable*/ ?

flatdata-cpp/test/GeneratedArchiveTest.cpp
```diff
@@ -330,6 +329,77 @@ TEMPLATE_TEST_CASE_METHOD( Fixture,
...
+    REQUIRE( outer.outer( ).value == 17u );
+
+    const char* const expected
+        = R"data(================================================================================
+Flatdata Archive: OuterWithOptional
+================================================================================
+Resource                             Optional  Too Large  Loaded    Details
+================================================================================
+outer                                NO        NO         YES       Structure of size 1
+archive_resource                     YES       NO         NO        FATAL: Resource storage not initialized. Please check archive path.
```
VeaaC, line 393: I guess the FATAL should be removed, since it is not fatal for optional subarchives

Same if true for `check archive path`.

Can you rephrase it into a neutral message? Or have a different message for root and nested?

PR #190, opened 2020-06-03
Update generator package to proper 0.4 release

VeaaC:
The 4.0-1 release way buggy: It did not contain a crucial Rust change


VeaaC:
Not needed once since #191 bumps version anyway

PR #191, opened 2020-06-19
Rust: Make archives Sync+Send

### LINKED ISSUE #165: The generated archive should be thread-safe
The rust implementation the `flatdata::ResourceStorage` is not thread safe and is missing a `std::marker::Send` implementation. 

There are usages of `RefCell` and `rc::Rc` which probably need to be assessed for thread safety.

VeaaC:
Fixes #165


VeaaC:
This also removes the Streams being `Rc<RefCell<T>>` for now reason except the memory resource storage needing it.

PR #195, opened 2020-10-07
Fix flatdata-inspector failing to open archive with enumeration fields

VeaaC:


flatdata-generator/flatdata/generator/generators/python.py
```diff
@@ -70,7 +70,10 @@ def to_dtype(flatdata_type):
                 "i32": "i4",
                 "u64": "u8",
                 "i64": "i8"
-            }[flatdata_type.name]
+            }
+            if field.type.name in type_map:
+                return type_map[field.type.name]
+            return type_map[field.type_reference.node.type.name]
```
gferon, line 76: you could have it in one line without the if:

```suggestion
            return type_map.get(field.type.name, field.type_refererence.node.type.name)
```

VeaaC: Isn't the 2nd parameter a default value, not a backup key?

Also: field.type_refererence is None if it is one of the base types.

gferon: Haha, yes, I done goofed. Ignore my comment, LGTM.

gferon (approved):

PR #196, opened 2020-10-09
Fix versioning hell

VeaaC:


boxdot:
Not easy to get the versioning right. Thanks for fixing this.

boxdot (approved):

PR #199, opened 2021-01-06
Fix default initializing ArrayView for overlapping structs

VeaaC:
Previously this would underflow and create a view for maximum possible size, starting from nullptr.


boxdot (approved):

gferon (commented):
commenting out the `nest_level` parameter in `MemoryDescriptor::describe` is a way to deprecate it without breaking backwards compatibility, right?

VeaaC:
@gferon It just marks the parameter as unused for this implementation of the resource interface

Issue #182, opened 2020-05-15
Add support for reading from TAR archive


VeaaC:
Flatdata archives are currently stored as dictionaries. For distribution it could be much nicer if they instead were a single file.

I propose that we should add support die the ResourceStorage(s) of the different languages read from (memory mapped) TAR archives

VeaaC:
Waiting for the `tar` crate to release the `Seek` improvements before closing this one.

PR #208, opened 2021-09-21
Add support for reading from TAR archives to flatdata-cpp (#182)

fermeise:


flatdata-cpp/include/flatdata/TarFileResourceStorage.h
```diff
@@ -0,0 +1,119 @@
...
+// -------------------------------------------------------------------------------------------------
+
+inline std::unique_ptr< TarFileResourceStorage >
+TarFileResourceStorage::create( const char* tar_path, const char* sub_path )
+{
+    std::shared_ptr< const MemoryMappedTarFileStorage > storage(
+        new MemoryMappedTarFileStorage( tar_path ) );
+
+    return std::unique_ptr< TarFileResourceStorage >(
+        new TarFileResourceStorage( storage, tar_path, sub_path ) );
```
VeaaC, line 62: What about error checking here (e.g. compare to FileResourceStorage which checks some things in `create`)

flatdata-cpp/src/TarReader.cpp
```diff
@@ -0,0 +1,192 @@
...
+/* Values used in typeflag field.  */
+const char REGTYPE = '0';           /* regular file */
+const char AREGTYPE = '\0';         /* regular file */
+const char DIRTYPE = '5';           /* directory */
+const char GNUTYPE_LONGNAME = 'L';  // Identifies the *next* file on the tape as having a long name.
+
+uint64_t
+decode_numeric_field( const char* data, size_t size )
+{
+    const uint64_t MAX_VALUE = std::numeric_limits< size_t >::max( );
```
VeaaC, line 54: assert that size != 0?

flatdata-cpp/src/TarReader.cpp
```diff
@@ -0,0 +1,192 @@
...
+
+    if ( *data & 0x80 )  // Highest byte set indicates base-256 encoding
+    {
+        uint64_t value = ( *data & 0x7F );
+        for ( size_t idx = 1; idx < size; ++idx )
+        {
+            if ( value > MAX_VALUE / 256 )
+            {
+                throw std::runtime_error( "Numeric value too large" );
```
VeaaC, line 63: This kind of error we should most likely catch in the `create` method: FileResourceStorage does not throw exceptions on opening (I think).

Also: Maybe some more context for the error message (error opening TAR file, etc)?

PPS: Flatdata is currently a bit bad on proper error feedback channels in C++ (Rust version is a bit better, but also a bit worse on some regards)

flatdata-cpp/src/TarReader.cpp
```diff
@@ -0,0 +1,192 @@
...
+}
+
+bool
+verify_checksum( const TarFileHeader& header )
+{
+    TarFileHeader header_copy = header;
+
+    memset( header_copy.chksum, ' ', sizeof( header_copy.chksum ) );
+    uint32_t sum = 0;
+    for ( size_t idx = 0; idx < sizeof( TarFileHeader ); ++idx )
```
VeaaC, line 94: `sizeof( header_copy )`?

flatdata-cpp/include/flatdata/MemoryMappedTarFileStorage.h
```diff
@@ -0,0 +1,75 @@
...
+        }
+
+        m_region = std::move( region );
+    }
+    catch ( boost::interprocess::interprocess_exception& )
+    {
+        return;
+    }
+
+    auto files = internal::read_tar_files( tar_path );
```
VeaaC, line 50: Why not pass the memory mapped region on to the TAR reader, instead of re-reading the data through conventional APIs? `read` would then just be a `memcpy`

flatdata-cpp/src/TarReader.cpp
fermeise: See above: What is the recommended way to propagate error messages in flatdata?

flatdata-cpp/include/flatdata/TarFileResourceStorage.h
VeaaC: I would not mind printing to std::cerr until we figure out how to properly propagate error messages.

VeaaC (approved):
Looks good to me.

PR #209, opened 2021-09-29
Add support for reading from TAR archives to flatdata-rs (#182)

fermeise:


flatdata-rs/lib/Cargo.toml
```diff
@@ -13,4 +13,5 @@ edition = "2018"
 [dependencies]
 diff = "0.1.11"
 memmap2 = "0.2.0"
+tar = "0.4"
```
VeaaC, line 16: Since this adds another dependency, can you make it an optional feature flag for the crate?

flatdata-rs/lib/src/tarstorage.rs
```diff
@@ -0,0 +1,128 @@
...
+#[derive(Debug)]
+struct MemoryMappedTarArchiveStorage {
+    archive_map: Mmap,
+    file_ranges: HashMap<PathBuf, Range<usize>>,
+}
+
+impl MemoryMappedTarArchiveStorage {
+    pub fn new(tar_path: &Path) -> Result<Self, io::Error> {
+        let file = File::open(tar_path)?;
+        let mut archive = tar::Archive::new(file);
```
VeaaC, line 24: Do you have to pass the file here, or can you also pass the memory mapped file? (I think you can convert a byte slice into Read/whatever trait is needed here, e.g. using `std::io::Cursor`)

flatdata-rs/lib/src/tarstorage.rs
```diff
@@ -0,0 +1,128 @@
...
+            .entries()?
+            .map(|entry| {
+                let entry = entry?;
+                let path = entry.path()?;
+                let path = if let Ok(stripped_path) = path.strip_prefix(".") {
+                    stripped_path.to_path_buf()
+                } else {
+                    path.to_path_buf()
+                };
+                let offset = entry.raw_file_position() as usize;
```
VeaaC, line 36: Is there a check anywhere that the file is actually stored in a consecutive way?

fermeise: There are AFAIK two cases in which a file is not stored contiguously:
1. Sparse files (files with holes in them)
2. Multivolume files

I added a check to fail in case we encounter a sparse file.
Multivolume files are currently not supported by tar-rs.

flatdata-rs/lib/src/tarstorage.rs
```diff
@@ -0,0 +1,133 @@
...
+            // (we never delete mappings), so we need to manually extend lifetime
+            let extended_lifetime_archive_map =
+                unsafe { slice::from_raw_parts(self.archive_map.as_ptr(), self.archive_map.len()) };
+
+            &extended_lifetime_archive_map[range.clone()]
+        })
+    }
+}
+
+/// Read-only resource storage on disk using a memory mapped tar archive.
```
VeaaC, line 69: You might want to reference this from the main lib.rs docs

flatdata-rs/tests/coappearances/Cargo.toml
```diff
@@ -10,3 +10,6 @@ flatdata = { path = "../../lib" }
 
 [build-dependencies]
 flatdata = { path = "../../lib" }
+
+[features]
+tar = ["flatdata/tar"]
```
VeaaC, line 15: EOL missing

VeaaC (approved):

PR #212, opened 2021-11-10
Check structs used in explicit references

### LINKED ISSUE #202: Flatdata schema parser does not check explicit_reference
It is currently possible to annotate a resource with a reference originating from a structure not present in said resource, e.g.:

```
namespace prime {
struct Factor {
    value : u32 : 32;
}
struct Number {
    @range(factors)
    first_factor_ref : u32;
}

archive Archive {
    @explicit_reference( Factor.value, factors ) // <-- bug / typo not caught
    numbers : vector<Number>;
    factors : vector<Factor>;
}
}

```

Expectation: Throw an error

VeaaC:
Generated error messages like:
`CRITICAL - app:71 - Error reading schema: Struct 'Factor' referenced, but not appearing in resource 'numbers'`

Fixes #202


eikesr (approved):
lgtm

PR #215, opened 2021-11-17
Add possibility to retrieve min and max for Struct fields in C++

### LINKED ISSUE #185: Expose maximum possible value for primitive types in generated code
While playing with flatdata, I realized that if you encode a primitive field with a custom size like so:

```
struct Attribute {
    value : u32 : 3;
}
```

then in the generated code, the number of bytes of the field is not exposed anywhere.

This could be useful for example when adding a `static_assert` in C++ that internal enum values match the size of the field. I suppose in Rust, we could also leverage this for some safer fun!

What I would have to do now:

```cpp
static_assert( ::MyEnum::COUNT == 8, "Oops, make sure your flatdata schema is up-to-date!" );
```

what I'd expect:

```cpp
static_assert( ::MyEnum::COUNT == something_max_value(), "..." );
```

pinging @boxdot and @VeaaC to discuss the proposal before I get to the implementation!

VeaaC:
Fixes #185

Min/max are lower_case for consistency's sake: bit_width already was

### DIFF
--- a/flatdata-cpp/include/flatdata/internal/Reader.h
+++ b/flatdata-cpp/include/flatdata/internal/Reader.h
@@ -30,17 +30,31 @@ template < typename T, int offset = 0, int num_bits = sizeof( T ) * 8, int struc
 struct Reader
 {
     using StreamType = const unsigned char*;
-    using UnsignedType = typename BitsToUnsigned<
-        int_choice< num_bits, num_bits + offset % 8, num_bits + offset % 8 <= 64 >::value >::type;
+
     enum
     {
         bit_width = num_bits
+    };
+    enum : typename UnderlyingType< T >::type
+    {
+        max = bit_width == 0 ? 0
+                             : ( bit_width - std::is_signed< T >::value
+                                         == sizeof( typename BitsToUnsigned< num_bits >::type ) * 8
+                                     ? typename BitsToUnsigned< num_bits >::type( -1 )
+                                     : ( typename BitsToUnsigned< num_bits >::type( 1 )
+                                         << ( bit_width - std::is_signed< T >::value ) )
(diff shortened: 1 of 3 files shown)


flatdata-cpp/include/flatdata/internal/Reader.h
```diff
@@ -30,17 +30,31 @@ template < typename T, int offset = 0, int num_bits = sizeof( T ) * 8, int struc
...
-        int_choice< num_bits, num_bits + offset % 8, num_bits + offset % 8 <= 64 >::value >::type;
+
     enum
     {
         bit_width = num_bits
     };
+    enum : typename UnderlyingType< T >::type
+    {
+        max = bit_width == 0 ? 0
+                             : ( bit_width - std::is_signed< T >::value
```
eikesr, line 41: Not the biggest fan of nesting ternary operators

eikesr (approved):

flatdata-cpp/include/flatdata/internal/Reader.h
VeaaC: Any suggestions for how to improve this?

eikesr: Is it possible to do as an immediately executed lambda?
```
max = []()
{
    if ( condition )
    {... return value;}
    else
    {... return value;}
}();
```

VeaaC: Not in C++11 (this needs to be constexpr)

eikesr: Hm, I can't think of anything else immediately

PR #217, opened 2021-11-19
Make it easier to use flatdata-rs generated code

### LINKED ISSUE #150: Simplify importing flatdata-rs
At the moment, to import flatdata in rust, we need to write:

```rust
#[macro_use]
extern crate flatdata
```

* We should try to remove this requirement.
* Also, nightly compiler fails due to missing `flatdata_intersperse` macro.

VeaaC:
Fixes #150


boxdot (approved):
Ah, this is very nice. I was wondering why we needed extern crate.

Issue #218, opened 2021-11-25
flatdata does not report errors in optional sub-archives


VeaaC:
Optional sub-archives in flatdata are not propagating serious errors when being opened. Instead serious errors are treated the same was as missing data.

PR #219, opened 2021-11-26
Rust: Handle errors in optional sub-archives

VeaaC:
Previously errors in optional sub-archive would be swallowed and the whole sub-archive just marked as missing.

Now flatdata-rs checks properly if the sub-archive is completely missing, or whether is has serious errors (broken resources, missing mandatory resources, etc)

Relates to #218


boxdot (approved):
Thank you for extending this!

flatdata-generator/flatdata/generator/templates/rust/archive.jinja2
```diff
@@ -153,23 +149,19 @@ impl {{archive.name}} {
                         data
                     ))
                 }
-                (Ok(_), Err(x)) | (Err(x), Ok(_)) => {return Err(x);}
-                (Err(x), Err(_)) => Err(x),
+                // is resource completely missing?
```
InfinityByTen, line 152: Minor: I'm wondering if the `let result` match could be a helper like `check`?

InfinityByTen (approved):

PR #222, opened 2021-12-16
Fix flatdata-rs release and adjust CI to check all features

VeaaC:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,5 @@
 sudo: required
-dist: bionic
+dist: focal
 language: minimal
 
 matrix:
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -112,6 +112,8 @@ matrix:
         - cd flatdata-rs
         - cargo build
         - cargo test
+        - cargo build --all-features
+        - cargo test --all-features
```
eikesr, line 116: Why do we still need l113-114? Isn't it enough to have the `cargo` commands with `--all-features`?

VeaaC: Yes: We want to make sure it compiles without the features enabled

eikesr (approved):

.travis.yml
boxdot: `cargo test` also does not link in contrary to `cargo build`

PR #223, opened 2022-01-12
Add Send/Sync annotation to allow ExternalVector to be send to other threads

VeaaC:

### DIFF
--- a/flatdata-rs/lib/src/storage.rs
+++ b/flatdata-rs/lib/src/storage.rs
@@ -14,7 +14,7 @@ use std::{
 };
 
 /// A handle to a resource storage used by archives
-pub type StorageHandle = Arc<dyn ResourceStorage + std::marker::Sync + std::marker::Send>;
+pub type StorageHandle = Arc<dyn ResourceStorage + Sync + Send>;
 
 pub trait Stream: Write + Seek {}
 impl<T: Write + Seek> Stream for T {}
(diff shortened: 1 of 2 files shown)


flatdata-rs/lib/src/storage.rs
```diff
@@ -13,8 +13,6 @@ use std::{
     sync::Arc,
 };
 
-use diff;
```
boxdot, line 16: Is diff still used? If not, we could remove it from toml?

flatdata-rs/lib/src/storage.rs
```diff
@@ -195,7 +193,7 @@ pub fn check_resource<T>(
 /// storage.
 #[doc(hidden)]
 pub fn create_external_vector<'a, T>(
-    storage: &'a dyn ResourceStorage,
+    storage: &'a (dyn ResourceStorage + std::marker::Sync + std::marker::Send),
```
boxdot, line 196: I think Sync and Send are in prelude, so technically no fqn is needed.

boxdot (approved):

flatdata-rs/lib/src/storage.rs
VeaaC: Still used, but directly (`dif::xyz`)

flatdata-rs/lib/src/storage.rs
VeaaC: Done

PR #227, opened 2022-07-18
Add support for user provided flatadata generator bin

gmarti:
Hello,

We are using nix to build our rust project and the flatdata generator installation script is not working well with nix reproducibility.

I think allowing the user to provide a path to the already build generator would be nice.

What do you think?

### DIFF
--- a/flatdata-rs/lib/src/generator.rs
+++ b/flatdata-rs/lib/src/generator.rs
@@ -59,6 +59,12 @@ use std::{
 /// If you are working on the generator, you can make sure your `build.rs`
 /// script picks up the source by setting `FLATDATA_GENERATOR_PATH` to point to
 /// the `flatdata-generator` folder.
+/// 
+/// ## Build
+/// 
+/// This method will try to install flatdata-generator in a python venv automatically
+/// You can provide your own generator by setting `FLATDATA_GENERATOR_BIN` to point to
+/// the `flatdata-generator` binary.
 pub fn generate(
     schemas_path: impl AsRef<Path>,
     out_dir: impl AsRef<Path>,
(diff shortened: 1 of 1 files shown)


VeaaC:
In general: We are always happy to get suggestions to improve the build / integration into build systems / etc (especially reproducibility).

Did you look into using the `FLATDATA_GENERATOR_PATH` environment variable (see https://docs.rs/flatdata/0.5.3/flatdata/fn.generate.html#development)? I have seen that used in integrated build system (which builds everything from source in one big build). We could also consider adding that as a function argument, but cargo invocations mostly forward data with environment variables, so it seemed a good fit.

flatdata-rs/lib/src/generator.rs
```diff
@@ -66,34 +66,42 @@ pub fn generate(
...
-        .wait()?;
-
-    // install dependencies
-    let generator_path = if let Ok(path) = env::var("FLATDATA_GENERATOR_PATH") {
-        // we want to rebuild automatically if we edit the generator's code
-        let path = PathBuf::from(path).canonicalize()?;
-        println!("cargo:rerun-if-changed={}", path.display());
-        eprintln!("installing flatdata-generator from source");
-        path
+    let generator_bin = if let Ok(bin_path) = env::var("FLATDATA_GENERATOR_BIN") {
```
VeaaC, line 69: Can you also update the docs of this function?

gmarti: Hello, i've added some documentation about FLATDATA_GENERATOR_BIN

flatdata-rs/lib/src/generator.rs
```diff
@@ -117,7 +131,8 @@ pub fn generate(
         );
 
         std::fs::create_dir_all(result.parent().unwrap())?;
-        let exit_code = Command::new(out_dir.join("bin/flatdata-generator"))
+
+        let exit_code = Command::new(generator_bin.clone())
```
VeaaC, line 135: Do you need a clone here?

flatdata-rs/lib/src/generator.rs
```diff
@@ -59,41 +59,55 @@ use std::{
...
+        PathBuf::from(bin_path)
     } else {
-        eprintln!("installing flatdata-generator from PyPI");
-        PathBuf::from("flatdata-generator==0.4.5")
+        // create a virtualenv in the target folder
+        eprintln!("creating python virtualenv");
+        let _ = Command::new("python3")
+            .arg("-m")
+            .arg("venv")
+            .arg("--system-site-packages")
```
VeaaC, line 84: Do you want to keep this change in? I am slightly concerned about this one, since Python packages notoriously sometimes have some conflicts between pip installed versions, and system packages.

flatdata-rs/lib/src/generator.rs
gmarti: i think so, since we are in a for loop

VeaaC (approved):
Looks good to me (FYI @boxdot)

VeaaC:
@gmarti thanks for this, looks ready to merge to me. The only thing missing is the DCO signoff ( see https://github.com/heremaps/flatdata/pull/227/checks?check_run_id=7410736588 / https://github.com/apps/dco )

PR #230, opened 2022-08-21
Fix (clippy) warnings in Rust generate code

boxdot:
Also fix (clippy) warnings in Rust tests

Tested against rustc version 1.63.0. After this is merged and the new
generator is released, additional flag `--all-targets` can be added to
`cargo test` in CI.


flatdata-generator/flatdata/generator/templates/rust/archive.jinja2
```diff
@@ -109,15 +109,13 @@ impl ::std::fmt::Debug for {{archive.name}} {
 }
 
 impl {{archive.name}} {
+    #[allow(unused_imports, unused_variables)]
     pub fn open(storage: flatdata::StorageHandle)
         -> ::std::result::Result<Self, flatdata::ResourceStorageError>
     {
-        #[allow(unused_imports)]
         use flatdata::SliceExt;
```
VeaaC, line 116: I remember these were added to make sure the "empty" archive does not generate errors (as it would not contain any code in here). Can you double-check that that works fine?

flatdata-generator/flatdata/generator/templates/rust/archive.jinja2
```diff
@@ -109,15 +109,13 @@ impl ::std::fmt::Debug for {{archive.name}} {
 }
 
 impl {{archive.name}} {
+    #[allow(unused_imports, unused_variables)]
```
VeaaC, line 112: This is a bit less granular, though, isn't it?

What's the problem with annotating exactly the imports/variable that might be unused?

flatdata-rs/lib/src/arrayview.rs
```diff
@@ -66,7 +66,7 @@ mod test {
         assert_eq!(vec.len(), 3);
         assert_eq!(vec[0].x(), 10..20);
         assert_eq!(vec[1].x(), 20..30);
-        assert_eq!(vec[2].x(), 30..0);
+        assert!(vec[2].x().is_empty());
```
VeaaC, line 69: This is not the same check, is it? A range can also be "non-nonsensical", that is going backwards

flatdata-rs/lib/src/arrayview.rs
```diff
@@ -99,22 +99,6 @@ mod test {
             assert_eq!(0, x.x());
             assert_eq!(0, x.y());
         }
-
```
VeaaC, line 102: Why is this removed?

flatdata-rs/lib/src/structs.rs
```diff
@@ -140,6 +144,6 @@ mod test {
 
     #[test]
     fn test_range() {
-        assert_eq!(<R as Struct>::IS_OVERLAPPING_WITH_NEXT, true);
+        const _: [(); 0 - !{ <R as Struct>::IS_OVERLAPPING_WITH_NEXT } as usize] = [];
```
VeaaC, line 147: What is this? :-)

flatdata-rs/lib/src/vector.rs
```diff
@@ -349,7 +349,7 @@ mod tests {
 
         assert_eq!(v[0].x(), 10..20);
         assert_eq!(v[1].x(), 20..30);
-        assert_eq!(v[2].x(), 30..0);
+        assert!(v[2].x().is_empty());
```
VeaaC, line 352: See previous comment, this does not test the same

flatdata-generator/flatdata/generator/templates/rust/archive.jinja2
boxdot: It did not work in the first place:

```
683 |         #[allow(unused_variables)]
     |         ^^^^^^^^^^^^^^^^^^^^^^^^^^ help: if you just forgot a `!`, use: `#![allow(unused_var
```

There is no way to annotated a single import as maybe being unused.

flatdata-rs/lib/src/structs.rs
boxdot: This is static assert: IS_OVERLAPPING_WITH_NEXT is actually just a constant and clippy complains that there is nothing to test. I added an allow directive and reverted it back.

flatdata-rs/lib/src/vector.rs
boxdot: Done.

VeaaC:
@boxdot btw, what's up with this one?

boxdot:
> @boxdot btw, what's up with this one?

Technically, I am still waiting on you review. The only one comment left, where we need to decide how to go forward.

flatdata-generator/flatdata/generator/templates/rust/archive.jinja2
VeaaC: But annotating a single variable is possible, should we do that?

PR #234, opened 2022-10-27
Fix dangling reference in ArrayViewIterator

VeaaC:
* Upgrade Catch as the old version is not compiling with new glibc anymore
* Upgrade to C++17 as Catch is not compiling with C++11 anymore


eikesr (approved):
lgtm

And interesting that catch is not header only anymore

PR #236, opened 2023-02-22
Improve flatdata-inspector

VeaaC:
* Do not copy memory when slicing RawData (massive speed gain on some scenarios)
* Support slicing Multivector
* Fix out-of-bounds for Multivector's len()
* add helper functions to extract strings from RawData, as well as string lists and fixed sized arrays
* Downgrade CI image since `nose` is not working with Python 3.10
* Simplify TAR backend a bit by also using memoryview

### DIFF
--- a/.github/workflows/generator.yml
+++ b/.github/workflows/generator.yml
@@ -11,7 +11,7 @@ env:
 
 jobs:
   Build:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-20.04
     steps:
       - uses: actions/checkout@v2
       - name: Install
--- a/flatdata-py/flatdata/lib/resources.py
+++ b/flatdata-py/flatdata/lib/resources.py
@@ -1,5 +1,5 @@
 '''
- Copyright (c) 2017 HERE Europe B.V.
+ Copyright (c) 2023 HERE Europe B.V.
  See the LICENSE file in the root of this project for license details.
 '''
 
--- a/flatdata-py/flatdata/lib/resources.py
+++ b/flatdata-py/flatdata/lib/resources.py
@@ -190,16 +190,16 @@ class RawData(ResourceBase):
 
     def sub_str(self, index, separator = b'\0'):
         for i in range(index, len(self)):
-            if all(a == b for (a, b) in zip(self[i:], separator)):
+            if self[i:i + len(separator)] == separator:
                 return bytes(self[index:i]).decode("utf-8")
         return bytes(self[index]).decode("utf-8")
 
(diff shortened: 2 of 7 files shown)


.github/workflows/generator.yml
```diff
@@ -11,7 +11,7 @@ env:
 
 jobs:
   Build:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-20.04
```
eikesr, line 14: Why not 22.04?

flatdata-py/flatdata/lib/resources.py
```diff
@@ -171,6 +188,32 @@ def __getitem__(self, item):
             ]
         return self._mem[item + SIZE_OFFSET_IN_BYTES:item + SIZE_OFFSET_IN_BYTES + 1]
 
+    def sub_str(self, index, separator = b'\0'):
+        for i in range(index, len(self)):
+            if all(a == b for (a, b) in zip(self[i:], separator)):
```
eikesr, line 193: Is `self[i:]` not a byte-string? I would find using `startswith` more readable, e.g.:

```
b'abc\0\0def'[3:].startswith(b'\0\0')
```

Or even scrap the loop and use `find`?

.github/workflows/generator.yml
VeaaC: 22.04 Has Python 3.10, and our current unit test framework (nose) does not work with Python 3.10 :-(

flatdata-py/flatdata/lib/resources.py
VeaaC: It is `bytes`, not a string. How would a version of `find` that gives me an index look like?

eikesr: https://docs.python.org/3/library/stdtypes.html#bytes.find would work? Although you need to check for `-1` as a return value in case the substring was not found

```
separator_index = self.find(separator, start=index)
# or
separator_index = self[index:].find(separator)
# then
if separator_index == -1:
  return bytes(self[index]).decode("utf-8")
return bytes(self[index:separator_index).decode("utf-8")
```

VeaaC: `AttributeError: 'memoryview' object has no attribute 'find'`

And converting to `bytes` reads all the data into memory

eikesr: Hmm, still find the zip solution a bit obscure, but I guess there is now easy way around it

eikesr (approved):

VeaaC:
Agree, but I could not find anything easy to use... I am not expert in Python, though

PR #242, opened 2024-04-24
Update generator to work with python 3.12

eikesr:
`imp` has been removed in python 3.12. I switched to the replacement suggested in the release notes. See: https://docs.python.org/3.12/whatsnew/3.12.html#imp

While I did not test previous python versions, the replacement function `types.ModuleType` is available since at least python 3.4.

Please have a look


eikesr:
Confirmed the pip parallel local build issue: https://github.com/pypa/pip/issues/8168

boxdot (approved):

eikesr:
Now worked around  the new setuptools+pip limitation (https://pip.pypa.io/en/stable/topics/local-project-installs/#build-artifacts) by switching to hatchling. Verified that the generated package is good (all files are present, seems to be usable).

One of you needs to upload the new version, once this is in

eikesr:
As for the other python package(s?) under `flatdata-py`. They are not affected, because they pull in `flatdata-generator` with pip and as such the build backend used for it doesn't matter.

I will, however, file a follow-up MR for upgrading the generator version used in the other packages. And maybe in that one change the build backend for them to hatchling, too, so that it is not a wild mix

VeaaC (approved):
LGT

PR #243, opened 2024-06-14
Move other py parts to hatchling; update generator version used

eikesr:
Move other py parts of flatdata, e.g., flatadata-inspector and flatdata-writer to hatchling, so that only one pip build backend is used.

Also upgrade the flatdata-generator version used to the new version. This requires also an update of these packages.


VeaaC:
The previous build was using local files for building, not released packages (otherwise you could not test non-backwards-compatible changes)

eikesr:
> The previous build was using local files for building, not released packages (otherwise you could not test non-backwards-compatible changes)... or wasn't it?

Yes, you were right. The job failed because the `requirements.txt` still mentioned the old version of the generator

VeaaC (approved):

PR #249, opened 2024-11-14
Add closing delimiter that was missing in rust generated code

muralisoundararajan:
# Why this change is required?

When we try adding index range in flatdata like below example and generate the rust code, the generated rust code is invalid.

```
struct Node {
    @range(edges_range)
    @optional( NO_EDGES_REF )
    first_edge_ref : u32;
}
```

When we use the generated code, compilation fails with below reason:

```
    |
483 |     pub fn edges_range(&self) -> std::ops::Range<Option<u32>> {
    |                                                               - closing delimiter possibly meant for this
...
486 |         let check = |x| Some(x).filter(|&x| x != super::test::NO_EDGES_REF;
    |                                       ^ unclosed delimiter
487 |         check(start)..check(end)
488 |     }
    |     ^ mismatched closing delimiter
```

# What this change does?

- Add closing delimiter to jinja template


muralisoundararajan:
@VeaaC agreed encoding identical start/end values  seems better approach in this case.

For the flatdata generator, it would be better to fail in the validation itself.  We could close this pull request

VeaaC:
See #250

PR #250, opened 2024-11-14
Report errors when trying to using optional ranges

VeaaC:
`@range` attribute cannot be combined with @optional: Users should store empty ranges instead (see #249).

This change make sure to report an error

```
14:30:42 - CRITICAL - app:71 - Error reading schema: @range cannot be combined with @optional, store empty ranges instead: edges_range 
```

For this schema:

```
namespace foo{
const u32 NO_EDGES_REF = 200;
struct Node {
    @range(edges_range)
    @optional( NO_EDGES_REF )
    first_edge_ref : u32;
}
}
```

### DIFF
--- a/flatdata-py/pyproject.toml
+++ b/flatdata-py/pyproject.toml
@@ -7,7 +7,6 @@ name = "flatdata-py"
 version = "0.4.7"
 description = "Python 3 implementation of Flatdata"
 readme = "README.md"
-license = ""
 authors = [
     { name = "Flatdata Developers" },
 ]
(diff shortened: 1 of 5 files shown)


flatdata-py/pyproject.toml
```diff
@@ -7,7 +7,6 @@ name = "flatdata-py"
 version = "0.4.7"
 description = "Python 3 implementation of Flatdata"
 readme = "README.md"
-license = ""
```
boxdot, line 10: TIL: https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#classifiers

boxdot (approved):

flatdata-py/pyproject.toml
VeaaC: Yes, we already have one a few lines below

PR #251, opened 2024-12-10
Type annotations for const and enum types

mohapatr3:


VeaaC (approved):
Looks good to me

flatdata-generator/flatdata/generator/tree/helpers/basictype.py
```diff
@@ -16,6 +16,18 @@ class BasicType:
...
+    _TYPE_ANNOTATION = {
+        "bool": "",
+        "u8": "",
+        "i8": "",
+        "u16": "",
+        "i16": "",
+        "u32": "UL",
+        "i32": "L",
+        "u64": "ULL",
+        "i64": "LL"
```
eikesr, line 28: I'm wondering if we need to use the PRIdx identifiers from https://en.cppreference.com/w/cpp/types/integer instead, since some machines might have different `unsigned long`?
Or am I worrying unnecessarily?

VeaaC: `PRId64` and friends are for printf... but I guess one could use `UINT64_C(0x123)`

I have never seen that in use, though, since `long long` guarantees at least 64 bits. `long` might be 32 or 64, but we should not care if its more than we need. We are assigning to a typed enumeration, so you will get a compiler warning anyway?

eikesr: Makes sense

PR #258, opened 2026-04-13
MINOR: adding Reader/Writer std::optional variants, same as boost::op…

epavese:


VeaaC:
Can you also add a test to `TEST_CASE( "Invalid values are handled", "[Struct]" )`?

You also need to sign the DCO (see https://github.com/heremaps/flatdata/pull/258/checks?check_run_id=71083619761)

VeaaC (approved):

PR #263, opened 2026-05-11
Adds `import "path/to/file.flatdata";` support to the schema language.

VeaaC:
Imports are resolved relative to the importing file. Diamond and cyclic imports are handled (deduplication by absolute path, depth-first traversal with visited set).

Code generation uses separate compilation for C++ and Rust: only types from the root file are emitted, with `#include` directives (C++) or `pub use super::..` re-exports (Rust) referencing sibling generated modules. Python/Dot/Flatdata generators remain monolithic — all imported types are inlined.

Generator CLI gains `--depfile` flag producing Makefile-style dependency output. The CMake macro uses `DEPFILE` on 3.20+; older versions fall back to depending only on the root schema file.

Changes:
- Grammar extended with `import_statement` rule
- New `importer.py`: `resolve_imports(path)` returns ordered resolved files + import metadata
- `builder.py`: `build_ast_from_file(path)` merges multi-file ASTs, tags nodes with `source_file` and `is_local`
- `SyntaxTree` extended with `imports`, `source_file_map`, `is_local_node()`
- `Engine.from_file(path)` class method for file-based compilation
- BaseGenerator: `filter_nodes()` / `get_import_directives()` hooks
- CppGenerator: filters to local nodes, emits `#include` for imports
- RustGenerator: `_import_reexports_for_namespace()` computes `super::` depth based on namespace nesting + directory traversal depth
- Writer/Inspector updated to use `Engine.from_file()`
- Removed dead `definition` parameter from `Structure/Enumeration/Constant/Archive.create()`


mohapatr3:
looks good to me!

